PR #15, opened 2025-05-16
Feature/faster cars from behind

DerMega:
An Overlay that show incoming faster Cars from Behind. 
It shows the Driver Name, the relative distance in s and as background color the car Class color
Pops up when they are in a 3s window.
Relies on the data from the Relatives Overlay. Therefore its lightweight.


src/frontend/components/FasterCarsFromBehind/FasterCarsFromBehind.stories.tsx
```diff
@@ -0,0 +1,36 @@
+import { Meta, StoryObj } from '@storybook/react';
+import { FasterCarsFromBehind } from './FasterCarsFromBehind';
+import { TelemetryDecorator } from '../../../../.storybook/telemetryDecorator';
+import { DynamicTelemetrySelector, TEST_DATA_DIRS } from '../Standings/DynamicTelemetrySelector';
```
tariknz:
```suggestion
import { DynamicTelemetrySelector } from '../Standings/DynamicTelemetrySelector';
```

src/app/storage/defaultDashboard.ts
```diff
@@ -69,5 +69,15 @@ export const defaultDashboard: DashboardLayout = {
         height: 330,
       },
     },
+    {
+      id: 'fastercarsfrombehind',
+      enabled: true,
```
tariknz: Should we set this to off for now until we add the settings page for it?
```suggestion
      enabled: false,
```

test-data/1733030013074/session.json
```diff
@@ -3426,7 +3426,7 @@
         "CarClassPowerAdjust": "0.000 %",
         "CarClassDryTireSetLimit": "0 %",
         "CarClassColor": 16767577,
-        "CarClassEstLapTime": 113.6302,
+        "CarClassEstLapTime": 126.6302,
```
tariknz, line 3429: Is this supposed to hopefully make it appear for the mock telemetry? Its still blank for me. The nearest car is 2.1s behind me.

tariknz: Also the car behind is actually a slower class

PR #17, opened 2025-05-16
Show track rubbered state in the Weather overlay

fgc:
I think this is nice to have, can confirm feelings of "slow track" on entering the session.
Could be elsewhere, and this is not strictly weather, but it´s "conditions"
I chose the icon to indicate "slipperyness", maybe there's a better one
I'm not sure about the formatting, those spans look sus to me, but CSS stumps me
Again this might be a toggle in the options, but can´t see anyone not wanting it

### DIFF
--- a/src/frontend/components/Weather/hooks/useTrackRubberedState.tsx
+++ b/src/frontend/components/Weather/hooks/useTrackRubberedState.tsx
@@ -1,0 +1,12 @@
+import { useSessionStore } from '@irdashies/context';
+import { useStore } from 'zustand';
+
+export const useTrackRubberedState = () => {
+  return useStore(
+    useSessionStore,
+    (state) =>
+      state.session?.SessionInfo?.Sessions?.find(
+        (session) => session.SessionNum === 0
+      )?.SessionTrackRubberState
+  );
+};
--- a/src/frontend/components/Weather/Weather.tsx
+++ b/src/frontend/components/Weather/Weather.tsx
@@ -3,7 +3,9 @@ import { useTrackTemperature } from './hooks/useTrackTemperature';
 import { useTrackWeather } from './hooks/useTrackWeather';
 import { WeatherTemp } from './WeatherTemp/WeatherTemp';
 import { WeatherTrackWetness } from './WeatherTrackWetness/WeatherTrackWetness';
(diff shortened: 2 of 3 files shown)


src/frontend/components/Weather/hooks/useTrackRubberedState.tsx
```diff
@@ -0,0 +1,12 @@
+import { useSessionStore } from '@irdashies/context';
+import { useStore } from 'zustand';
+
+export const useTrackRubberedState = () => {
+  return useStore(
+    useSessionStore,
+    (state) =>
+      state.session?.SessionInfo?.Sessions?.find(
+        (session) => session.SessionNum === 0
```
xikxp1, line 9: Why is it SessionNum 0? I think you should use current session number. Like for example here: https://github.com/tariknz/irdashies/blob/76b1ba72a8d829b4a3db9cb9175556e68b431eae/src/frontend/components/Standings/hooks/useSessionLapCount.tsx#L5-L6C23

tariknz: That's a good point sorry I missed that

fgc: My fault for submitting unread code ;) I guess the effect is that you get stale track state for practice during qualy and race. I'll try to fix and submit a new PR later today when I get home

fgc: I had a look at this, and the plot seriously thickened:
- State of every session can be set specifically or "carry over" from the previous one, the state then is literally "carry over"
- Officials are set as "carry over", AI and hosted can be set to specific states per session or carry over.
- I think it's better to know the initial state than to show "carry over".
- So as it is now it works perfectly for official races.
- To make it work for hosteds we would need logic to show the "last non carry over state". But this is not as trivial as just memoizing the latest valid state and showing the latest non-carry one, because what happens to someone who joins in qualy? we need to look back into past sessions to find a valid one.

I almost regret submitting this now 🤣

fgc: I'm stupid it´s a very easy fix, I'll submit another PR with a sane handling of this

PR #66, opened 2025-08-28
Settings: show overlays only when on track

aalyokhin:
SCR-20250828-spau

With this setting enabled, overlays will be rendered only when the car is on the track or in the pit stall. Benefits:

- prevent overlays from showing up when iRacing is still being loaded
- rendering them when in the menus may cover important info on the screen
- no more need to close the app if recording the replay as a video

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -7,6 +7,7 @@ import {
   RunningStateProvider,
   useRunningState,
   SessionProvider,
+  useDrivingState,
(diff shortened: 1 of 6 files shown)


tariknz:
Are you on discord @aalyokhin ?

aalyokhin:
> Are you on discord @aalyokhin ?

@tariknz yeah, I've just changed the name coz had a nickname there

PR #82, opened 2025-11-06
feat: relative/standings car manufacturers

creagel:
Implemented the display of the vehicle brand logo (standing + relative).
Implemented multi-dry tires for supported vehicles.

image

image

### DIFF
--- a/src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
+++ b/src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
@@ -1,4 +1,3 @@
-import React from 'react';
 import carLogoImage from '../../../../assets/img/car_manufacturer.png';
 
 interface CarManufacturerProps {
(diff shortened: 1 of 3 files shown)


src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
```diff
@@ -0,0 +1,229 @@
...
+  117: 'holden',    // Supercars Holden ZB Commodore (legacy)
+  118: 'ford',      // Supercars Ford Mustang GT (legacy)
+  119: 'porsche',   // Porsche 718 Cayman GT4 Clubsport
+  120: 'unknown',   // Indy Pro 2000 PM-18
+  121: 'unknown',   // USF 2000
+  122: 'bmw',       // BMW M4 F82 GT4 - 2018
+  123: 'ford',      // NASCAR Truck Ford F150
+  124: 'chevrolet', //NASCAR Legends Chevrolet Monte Carlo - 1987
+  125: 'ford',      //NASCAR Legends Ford Thunderbird - 1987
+  127: 'chervrolet',// Chevrolet Corvette C8.R GTE
```
tariknz:
```suggestion
  127: 'chevrolet',// Chevrolet Corvette C8.R GTE
```

src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
```diff
@@ -0,0 +1,229 @@
...
+  "toyota": { x: '0', y: '61.50%' },
+  "holden": { x: '0', y: '64.12%' },
+  "nissan": { x: '0', y: '66.67%' },
+  "subaru": { x: '0', y: '69.22%' },
+  "hyundai": { x: '0', y: '71.77%' },
+  "ligier": { x: '0', y: '74.32%' },
+  "renault": { x: '0', y: '76.94%' },
+  "superformula": { x: '0', y: '79.51%' },
+  "dallara": { x: '0', y: '82.0%' },
+  "mercedes": { x: '0', y: '84.60%' },
```
tariknz: 'hpd', 'buick', and 'srx'  are missing - in those instances what would happen?

I suspect its a runtime error because `carLogoPositions[carManufacturer].x` would fail

src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
```diff
@@ -0,0 +1,229 @@
...
+};
+
+// Cars models
+export const CAR_ID_TO_CAR_MANUFACTURER: Record<number, string> = {
```
tariknz, line 57: Better type safety
```suggestion
export const CAR_ID_TO_CAR_MANUFACTURER: Record<number, keyof typeof carLogoPositions> = {
```

src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
```diff
@@ -0,0 +1,229 @@
...
+
+  if (carId < 0) {
+    return null;
+  }
+
+
+  let carManufacturer = getCarManufacturerFromCarID(carId);
+
+  if (!carManufacturer) {
+   carManufacturer = 'unknown';
```
tariknz: Can unknown just be a blank logo instead of that car with the `?`

src/frontend/components/Standings/components/DriverInfoRow/DriverInfoRow.tsx
```diff
@@ -131,11 +135,9 @@ export const DriverInfoRow = ({
...
         </td>
       )}
-     {tireCompound !== undefined && (
-        <td>
-          <div className="flex items-center pr-1"> 
-           <Compound tireCompound={tireCompound} size="sm" />
-          </div>
+     {tireCompound !== undefined && carId && (
+        <td className="px-2">
+           <Compound tireCompound={tireCompound} carId={carId} size="sm" className="mr-2 flex-shrink-0" />
```
tariknz: Having the container lets it be centred better vertically so let's keep that. It might've been overwritten when rebasing.
```suggestion
       <div className="flex items-center pr-1"> 
           <Compound tireCompound={tireCompound} carId={carId} size="sm" />
        </div>   
```

src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
```diff
@@ -0,0 +1,229 @@
...
+                                                    className = '',
+                                                    size = 'md',
+                                                  }) => {
+
+  if (carId < 0) {
+    return null;
+  }
+
+
+  let carManufacturer = getCarManufacturerFromCarID(carId);
```
tariknz: Just change to this
```suggestion
const carManufacturer = CAR_ID_TO_CAR_MANUFACTURER[carId] || 'unknown';
```

src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
```diff
@@ -0,0 +1,229 @@
...
+export const CarManufacturer: React.FC<CarManufacturerProps> = ({
+                                                    carId,
+                                                    className = '',
+                                                    size = 'md',
+                                                  }) => {
+
+  if (carId < 0) {
+    return null;
+  }
+
```
tariknz:
```suggestion
```

src/frontend/assets/img/car_manufacturer.png
tariknz, line 1: I'm still not quite sold on the sprite approach. So the main issue is that if someone wants to come add a new logo how do we support that? how do you make sure everything is in the right position? Its a little bit of a headache to maintain long term.

We can deal with it later but I think the performance savings is neglegable since its all local.

src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
creagel: Missing manufacturers was mistake. Added check for existing key in array.

src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
creagel: I don't think so - this is just a mapping of the vehicle ID to the manufacturer name (string key in another field)

src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
creagel: I removed the logo from the png file. We'll leave the mapping. In the future (if there will be a request) just edit the png file with the logo for the unknown manufacturer and it will be displayed.

src/frontend/assets/img/car_manufacturer.png
creagel: I leave free positions in sprint files (files are always bigger and there is room for the future). Yes, if the positions are exhausted, it is necessary to enlarge the file and "readdress" all current icons.

Especially in the case of manufacturers, I think that reloading (especially if a separate overlay for streaming is introduced) is definitely a benefit.

src/frontend/components/Standings/components/CarManufacturer/CarManufacturer.tsx
tariknz: That's fine but the idea is that ensures you don't use a string that's not mapped by accident.

PR #86, opened 2025-11-09
feat: add lap time delta columns to standings overlay

23Kev:
- Add configurable lap time delta columns (1-5 laps)
- Compare each driver's recent lap times to player's recent lap times
- Green text for positive deltas (driver slower), red for negative (driver faster)
- Player row displays '-' for delta columns
- Add settings toggle and lap count selector
- Centralize LapTimesStore updates via LapTimesStoreUpdater provider
- Optimize performance with version-based equality checking for endurance races
- Refactor useDriverStandings to accept settings object directly

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -13,7 +13,6 @@ import { Settings } from './components/Settings/Settings';
 import { EditMode } from './components/EditMode/EditMode';
 import { ThemeManager } from './components/ThemeManager/ThemeManager';
 import { WIDGET_MAP } from './WidgetIndex';
-import { LapTimesStoreUpdater } from './context/LapTimesStore/LapTimesStoreUpdater';
 
 const AppRoutes = () => {
   const { currentDashboard } = useDashboard();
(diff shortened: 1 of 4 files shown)


src/frontend/App.tsx
```diff
@@ -55,13 +56,15 @@ const App = () => (
...
       <SessionProvider bridge={window.irsdkBridge} />
       <TelemetryProvider bridge={window.irsdkBridge} />
-      <HashRouter>
-        <EditMode>
-          <ThemeManager>
-            <AppRoutes />
-          </ThemeManager>
-        </EditMode>
-      </HashRouter>
+      <LapTimesStoreUpdater>
```
tariknz: I think we should only add this provider for the standings as that means we'll be tracking relative for every overlay.

The way it works currently unfortunately is because every overlay is a seperate browser window that means each overlay has independent set of providers given.

A better approach would just be to put it inside Standings.tsx or a container for it. It doesn't need to wrap its children or even be a component in this case just make it a hook.

23Kev: That makes sense, I'll make that update.

src/frontend/context/LapTimesStore/LapTimesStore.tsx
```diff
@@ -5,6 +5,7 @@ export interface LapTimeBuffer {
   lastLapTimes: number[];
   lastSessionTime: number;
   lapTimeHistory: number[][]; // [carIdx][sample]
+  version: number; // Incremented when lapTimeHistory changes
 }
 
 interface LapTimesState {
```
tariknz, line 11: I think we might need to add something to reset the state when the session ID changes. I don't remember if its been implemented

23Kev: Thats a good point, I do think I saw in testing the deltas carried over to the next session.
Will it be to reset the state for the whole app or just in the LapTimeStore?

tariknz: Just the LapTimeStore for now although we should ideally reset for the whole app but we can do that separately (just to avoid unexpected regression).

src/frontend/App.tsx
23Kev: I've made this change now, tested all fine.

src/frontend/context/LapTimesStore/LapTimesStore.tsx
23Kev: LapTimeStore is reset to 0 at the end of session or between sessions.

PR #87, opened 2025-11-11
feat: custom highlight color

LanceVShoot:
add custom highlight color to change the accent color of relatives/standings during single-class races. (iRacing multi-class colors used in multi-class races.) Updated first class color to match iracing (yellow instead of orange).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,6 +104,3 @@ src/**/tracks/tracks.json
 build/
 bin/
 docs/features/
-
-#package-lock
-package-lock.json
--- a/src/frontend/components/Standings/Relative.tsx
+++ b/src/frontend/components/Standings/Relative.tsx
@@ -76,7 +76,7 @@ export const Relative = () => {
         />
       );
     });
-  }, [standings, playerIndex, totalRows, config]);
+  }, [standings, playerIndex, totalRows, config, isMultiClass]);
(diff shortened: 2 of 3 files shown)


src/frontend/components/Standings/Relative.tsx
tariknz:
```suggestion
  }, [standings, playerIndex, totalRows, config, isMultiClass]);
```

src/types/dashboardLayout.ts
```diff
@@ -27,6 +27,7 @@ export interface GeneralSettingsType {
   fontSize?: 'xs' | 'sm' | 'lg' | 'xl';
   colorPalette?: 'default' | string;
   showOnlyWhenOnTrack?: boolean;
+  highlightColor?: 960745 | number;
```
tariknz:
```suggestion
  highlightColor?: number;
```
This is probably unecessary

.gitignore
```diff
@@ -104,3 +104,6 @@ src/**/tracks/tracks.json
 build/
 bin/
 docs/features/
+
+#package-lock
+package-lock.json
```
tariknz:
```suggestion
```
You shouldn't ignore package-lock that is what ci uses to install the specific versions.

PR #94, opened 2025-11-14
feat: standings/relative reorder columns in settings

LanceVShoot:
Thoughts on this? Main change is updating relatives to have the same settings/options as standings for consistency, and allowing columns to be re-ordered for custom display. Added settings window position saving so it reopens in the same location from when it was closed.


tariknz:
@LanceVShoot it seems like the registery for that package isn't from npm? 
Its coming from: https://onediversified.jfrog.io

Is that possibly a work one that caches your packages?

tariknz:
Is this still in draft? want me to look at it yet?

src/frontend/components/Standings/components/DriverInfoRow/DriverInfoRow.tsx
```diff
@@ -54,28 +58,235 @@ export const DriverInfoRow = memo(({
...
+      },
+      {
+        id: 'badge',
+        shouldRender: (displayOrder ? displayOrder.includes('badge') : true) && (config?.badge?.enabled ?? true),
+        render: () => <td key="badge" data-column="badge" className="whitespace-nowrap text-center">{hidden ? null : badge}</td>,
+      },
+      {
+        id: 'iRating',
+        shouldRender: (displayOrder ? displayOrder.includes('iRating') : true) && (config?.iRating?.enabled ?? false),
+        render: () => <td key="iRating" data-column="iRating" className="px-2 text-center just whitespace-nowrap">{hidden ? null : iRating}</td>,
```
tariknz:
```suggestion
        render: () => <td key="iRating" data-column="iRating" className="px-2 text-center whitespace-nowrap">{hidden ? null : iRating}</td>,
```
Not a valid css class - not sure if you intended it to be justify-center

src/frontend/components/Standings/components/DriverInfoRow/DriverInfoRow.tsx
```diff
@@ -54,28 +58,235 @@ export const DriverInfoRow = memo(({
...
+      return [...orderedColumns, ...remainingColumns];
+    }
+
+    // If no displayOrder, return all columns that should render in original order
+    return columns.filter(col => col.shouldRender);
+  }, [
+    carNumber, flairId, carId, badge, iRating, iratingChange, delta,
+    tireCompound, lapTimeDeltas, numLapDeltasToShow, isPlayer, displayOrder, config,
+    classColor, highlightColor, isMultiClass, hasFastestTime, lastTimeState,
+    fastestTimeString, lastTimeString, name, onPitRoad, position, radioActive, hidden
```
tariknz: This actually might be problamatic - that means every component will be rerendering when any of the items in the dependency change. We could use memo to each one under the render function. 

I'm not sure but we can address it later, worth testing.

src/frontend/components/Standings/Relative.tsx
```diff
@@ -27,7 +28,39 @@ export const Relative = () => {
...
+          isMultiClass={false}
+          lapTimeDeltas={undefined}
+          numLapDeltasToShow={undefined}
+          displayOrder={settings?.displayOrder}
+          config={settings}
+          carNumber={settings?.carNumber?.enabled ?? true ? '' : undefined}
+          flairId={settings?.countryFlags?.enabled ?? true ? 0 : undefined}
+          carId={settings?.carManufacturer?.enabled ?? true ? 0 : undefined}
+          badge={settings?.badge?.enabled ? <></> : undefined}
+          iRating={settings?.iRating?.enabled ? undefined : undefined}
```
tariknz: @LanceVShoot what was the reason for adding iRating on its own? its part of the badge. If you wanted to hide it seperately we can just make it a setting for the badge itself.

tariknz: Its currently enabled by default for standings

Image

LanceVShoot: I want to say it was in there somewhere? In standings or relatives settings? I know I didn't add it from the standpoint of me wanting to see it - it doesn't make sense to me either to show it when it's in the badge. Maybe I accidentally used iRating somewhere instead of iratingChange, and then got confused thinking it was there as an option. lol I'll remove it - I double checked and see it wasn't in the current main so it had to be me that added it by accident!

src/frontend/components/Settings/sections/RelativeSettings.tsx
```diff
@@ -3,18 +3,88 @@ import { BaseSettingsSection } from '../components/BaseSettingsSection';
...
+  { id: 'pitStatus', label: 'Pit Status', configKey: 'pitStatus' },
+  { id: 'carManufacturer', label: 'Car Manufacturer', configKey: 'carManufacturer' },
+  { id: 'badge', label: 'Driver Badge', configKey: 'badge' },
+  { id: 'iRating', label: 'iRating', configKey: 'iRating' },
+  { id: 'iratingChange', label: 'iRating Change', configKey: 'iratingChange' },
+  { id: 'delta', label: 'Delta', configKey: 'delta' },
+  { id: 'fastestTime', label: 'Best Time', configKey: 'fastestTime' },
+  { id: 'lastTime', label: 'Last Time', configKey: 'lastTime' },
+  { id: 'compound', label: 'Tire Compound', configKey: 'compound' },
+  { id: 'lapTimeDeltas', label: 'Lap Time Deltas', configKey: 'lapTimeDeltas', hasSubSetting: true },
```
tariknz: Can you not include the lapTimeDeltas for the relative - this won't work here because it requires a seperate store to be included. 

Right now for the relative blackbox I am worried about the performance overhead as this re-renders a lot more frequently than standings.

src/frontend/components/Standings/Relative.tsx
tariknz: Yep i see it might've not been a "displayable" field. Probably hard to distinguish them hah. Thanks!

src/frontend/components/Settings/sections/RelativeSettings.tsx
LanceVShoot: Absolutely! I was trying to standardize it so both boxes could display any setting so people could customize it as they wish. But makes total sense to leave that piece out - thanks for the heads up!

tariknz: Yep I totally get you - I feel the same. Its just this particular one is implemented in a way that if its on or off it still does its calculations internally. Just needs a little tidy up.

tariknz (approved):
LGTM

PR #98, opened 2025-11-17
feat: add enhanced relative gap

23Kev:
Now this should be a clean branch/PR for the Relative Gap Improvements feature that only contains the Relative files. 
The Brake Bias change is also included in here.

### DIFF
--- a/src/frontend/components/Standings/hooks/useDriverRelatives.tsx
+++ b/src/frontend/components/Standings/hooks/useDriverRelatives.tsx
@@ -4,9 +4,10 @@ import {
   useSessionStore,
   useTelemetryValues,
   useRelativeGapStore,
+  detectEdgeCases,
+  calculateRelativeGap,
 } from '@irdashies/context';
 import { useDriverStandings } from './useDriverPositions';
-import { detectEdgeCases, calculateRelativeGap } from '@irdashies/context/RelativeGapStore';
 import type { Standings } from '../createStandings';
 
 export const useDriverRelatives = ({ buffer }: { buffer: number }) => {
--- a/src/frontend/components/Standings/hooks/useDriverRelatives.spec.tsx
+++ b/src/frontend/components/Standings/hooks/useDriverRelatives.spec.tsx
@@ -12,6 +12,8 @@ vi.mock('@irdashies/context', async (importOriginal) => {
     useTelemetryValues: vi.fn(),
     useSessionStore: vi.fn(),
     useRelativeGapStore: vi.fn(),
+    detectEdgeCases: vi.fn(() => ({ isLapping: false, isBeingLapped: false, isMultiClass: false })),
+    calculateRelativeGap: vi.fn(() => null),
   };
 });
 
@@ -19,10 +21,8 @@ vi.mock('./useDriverPositions', () => ({
   useDriverStandings: vi.fn(),
 }));
 
-vi.mock('@irdashies/context/RelativeGapStore', () => ({
-  detectEdgeCases: vi.fn(() => ({ isLapping: false, isBeingLapped: false, isMultiClass: false })),
-  calculateRelativeGap: vi.fn(() => null),
-}));
+// Note: detectEdgeCases and calculateRelativeGap are now exported from @irdashies/context
+// We need to mock them in the context mock above
 
 // Import mocked functions after vi.mock
 const { useDriverCarIdx, useTelemetryValues, useSessionStore, useRelativeGapStore } = await import('@irdashies/context');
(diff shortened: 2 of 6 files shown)


src/frontend/context/RelativeGapStore/RelativeGapStoreUpdater.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { useSessionStore } from '../SessionStore/SessionStore';
+import { useRelativeGapStore } from './RelativeGapStore';
+
+/**
+ * Hook that automatically updates the RelativeGapStore with telemetry data.
+ * This samples position data and detects lap completions to build position/time records.
+ */
+export const useRelativeGapStoreUpdater = () => {
+  const telemetry = useTelemetryStore((state) => state.telemetry);
+  const session = useSessionStore((state) => state.session);
```
tariknz: This is not a good idea - you're going to get an update for both of these at 30hz
Just useTelemetryValue and for session there's some default helpers to pull what you need out. This will ensure we only update the store when the dependent telemetry values actually change.

src/frontend/context/RelativeGapStore/RelativeGapStoreUpdater.tsx
```diff
@@ -0,0 +1,85 @@
...
+    for (let carIdx = 0; carIdx < carIdxLapDistPct.length; carIdx++) {
+      const position = carIdxLapDistPct[carIdx];
+      const currentLap = carIdxLap[carIdx];
+      const lastLapTime = carIdxLastLapTime[carIdx];
+
+      // Skip invalid positions
+      if (position < 0 || currentLap < 0) continue;
+
+      // Initialize car history if needed
+      store.initializeCarHistory(carIdx, currentLap, sessionTime);
```
tariknz: Similar to that last PR you need to get these functions out of the store

src/frontend/context/RelativeGapStore/RelativeGapStoreUpdater.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+  // Track last lap numbers to detect lap completions
+  const lastLapNumbers = useRef<Map<number, number>>(new Map());
+
+  useEffect(() => {
+    if (!telemetry || !session) return;
+
+    const sessionTime = telemetry.SessionTime?.value?.[0] ?? 0;
+    const sessionNum = telemetry.SessionNum?.value?.[0] ?? -1;
+    const trackLength = parseTrackLength(session.WeekendInfo?.TrackLength);
```
tariknz: use the useTrackLength (in SessionStore.tsx) hook instead, that will give you the parsed length in meters. It will also be memoized so you don't need to rerun thsi function.

src/frontend/components/Standings/hooks/useDriverRelatives.tsx
```diff
@@ -3,19 +3,30 @@ import {
   useDriverCarIdx,
   useSessionStore,
   useTelemetryValues,
+  useRelativeGapStore,
 } from '@irdashies/context';
 import { useDriverStandings } from './useDriverPositions';
+import { detectEdgeCases, calculateRelativeGap } from '@irdashies/context/RelativeGapStore';
```
tariknz: export RelativeGapStore in the index.ts

src/frontend/context/RelativeGapStore/RelativeGapStoreUpdater.tsx
tariknz: Have a look at this when you get a moment
https://github.com/tariknz/irdashies/pull/104/files#diff-1f0584807de80ee7f361fdf60ee55e0ed96fe7a294c6e782c84763783545079f

Specifically the useLapTimesStoreUpdater - this is what I want to change around that.

23Kev:
Changes Made:
1. RelativeGapStore.tsx - Added `processPositionUpdates` action
Moved all business logic from the updater hook into the store
Created a single action method that processes all position updates
Follows the established pattern from `LapTimesStore`
2. RelativeGapStoreUpdater.tsx - Complete refactoring
- Replaced `useTelemetryStore((state) => state.telemetry)` with granular hooks:
```
useTelemetryValue('SessionTime')
useTelemetryValue('SessionNum')
useTelemetryValues('CarIdxLapDistPct')
useTelemetryValues('CarIdxLap')
useTelemetryValues('CarIdxLastLapTime')
```
Replaced `parseTrackLength()` function with `useTrackLength()` hook
Removed 18 lines of duplicate track length parsing code
Moved all business logic to store's `processPositionUpdates` actionHook is now minimal (15 lines vs 62 lines) - just passes data to store

PR #100, opened 2025-11-17
feat: additional font sizes

LanceVShoot:
Added additional larger font sizes and changed to a slider bar for selection


tariknz (approved):
nice one
LGTM

PR #105, opened 2025-11-18
Expose components through uri

DPShillcock:
# Component URI Exposure Feature

## Overview

This feature enables external browsers to access individual dashboard components via HTTP URLs, allowing components to be embedded in iframes, OBS browser sources, or viewed standalone in any web browser.

### Component Server (Port 3000)
- **Express HTTP Server**: Serves component pages at `http://localhost:3000/component/<name>`
- **WebSocket Bridge**: Real-time data streaming via Socket.io
  - Broadcasts telemetry (60fps), session data, and running state
  - Syncs dashboard configuration and demo mode status
  - Supports multiple concurrent browser connections

### Data Flow
```
iRacing SDK → Electron Main Process → WebSocket Server → Browser Clients
                     ↓
              Demo Mode Toggle
                     ↓
           Mock Data Generator (with multi-subscriber support)
```

### Key Components

1. **Bridge Proxy** (`bridgeProxy.ts`)
   - WebSocket server that exposes Electron bridge data
   - Tracks demo mode state and syncs to connected clients
   - Resubscribes when bridge changes (e.g., toggling between live/demo data)

2. **Component Renderer** (`componentRenderer.tsx`)
   - Browser-side React renderer with WebSocket client
   - Applies theme settings and dashboard configuration
   - Shows demo mode badge when using mock data
   - Supports conditional debug logging via `?debug=true`

3. **Mock Data Generator** (`generateMockData.ts`)
   - Refactored to support multiple subscribers using Sets
   - Generates 60fps telemetry and 2s session updates
   - Allows both overlay manager and browser clients to receive data

### Accessing Components

**Basic URL Pattern:**
```
http://localhost:3000/component/<component-name>
```

**Available Components:**
- `map` - Track map with driver positions
- `standings` - Race standings table
- `input` - Driver input visualization
- `fuel` - Fuel calculator
- `weather` - Weather information
- And all other dashboard widgets

**URL Parameters:**
- `debug=true` - Enable verbose console logging
- `config=<json>` - Override component configuration (URL-encoded JSON)

### Examples

**Track Map (Basic):**
```
http://localhost:3000/component/map
```

**Track Map (With Config):**
```
http://localhost:3000/component/map?config=%7B%22enableTurnNames%22%3Afalse%7D
```

**Standings (Debug Mode):**
```
http://localhost:3000/component/standings?debug=true
```

### Embedding in OBS

1. Add **Browser Source** to your scene
2. Set URL to: `http://localhost:3000/component/<name>`
3. Set dimensions (e.g., 1920x1080)
4. Enable "Shutdown source when not visible" for performance
5. Background is transparent by default

### Embedding in iframes

```html
<iframe 
  src="http://localhost:3000/component/map?config=%7B%22enableTurnNames%22%3Afalse%7D" 
  width="1920" 
  height="1080"
  scrolling="no"
  frameborder="0">
</iframe>
```

## Demo Mode

Toggle between live iRacing data and mock data:
- Use the Electron app's demo mode toggle
- Demo mode badge appears in browser components
- WebSocket automatically syncs state to all connected clients
- Bridge resubscribes when switching data sources

## Features

✅ **Real-time Data Streaming** - 60fps telemetry via WebSocket  
✅ **Transparent Backgrounds** - Ready for overlays and embedding  
✅ **Theme Support** - Respects dashboard font size and color settings  
✅ **Multi-Client** - Multiple browsers can connect simultaneously  
✅ **Demo Mode Sync** - Live indication when using mock data  
✅ **Debug Mode** - Conditional logging with `?debug=true`  
✅ **Hot Reload** - Components update when dashboard settings change  

### CSS Variables
- Uses Tailwind CSS v4 with custom `@theme` configuration
- `--spacing: 0.25rem` for consistent padding calculations
- Transparent backgrounds with `!important` to override dark color-scheme defaults

### Debug Logging
- Suppressed by default for clean console
- Enable with `?debug=true` URL parameter
- Shows WebSocket connection, data flow, and render lifecycle

### Browser Compatibility
- Modern browsers with WebSocket support
- Tested on Chrome, Edge, Firefox
- Works in OBS browser sources (CEF)

## Performance

- WebSocket maintains single connection per client
- Mock data generator uses single interval per data type
- Set-based subscriber pattern prevents callback overwrites
- 60fps telemetry with throttled session updates (2s)

### DIFF
--- a/src/app/webserver/bridgeProxy.ts
+++ b/src/app/webserver/bridgeProxy.ts
@@ -2,7 +2,7 @@ import { Server as HTTPServer } from 'http';
 import { WebSocketServer, WebSocket } from 'ws';
 import type { Telemetry, Session, DashboardLayout } from '@irdashies/types';
 import type { IrSdkBridge, DashboardBridge } from '@irdashies/types';
-import { getIsDemoMode } from './iracingSdk/setup';
+import { getIsDemoMode } from '../bridge/iracingSdk/setup';
 
 // Export current state so it can be accessed by other parts of the app
 export let currentDashboard: DashboardLayout | null = null;
@@ -156,3 +156,4 @@ export function createBridgeProxy(
 
   return { wss, resubscribeToBridge };
 }
+
--- a/src/app/webserver/componentRenderer.tsx
+++ b/src/app/webserver/componentRenderer.tsx
@@ -596,3 +596,4 @@ declare global {
 if (typeof window !== 'undefined') {
   (window as any).renderComponent = renderComponent;
 }
+
--- a/src/app/webserver/componentServer.ts
+++ b/src/app/webserver/componentServer.ts
@@ -210,7 +210,7 @@ export async function startComponentServer(irsdkBridge?: IrSdkBridge, dashboardB
       const { createBridgeProxy } = await import('./bridgeProxy');
       const { resubscribeToBridge } = createBridgeProxy(httpServer, irsdkBridge, dashboardBridge);
       
-      const { onBridgeChanged } = await import('./iracingSdk/setup');
+      const { onBridgeChanged } = await import('../bridge/iracingSdk/setup');
       onBridgeChanged((newBridge) => {
         resubscribeToBridge(newBridge);
       });
@@ -223,3 +223,4 @@ export async function startComponentServer(irsdkBridge?: IrSdkBridge, dashboardB
     console.log(`Component server running on http://localhost:${COMPONENT_PORT}`);
   });
 }
+
--- a/src/component-renderer-entry.ts
+++ b/src/component-renderer-entry.ts
@@ -7,7 +7,7 @@
 // Only import what we need for the browser environment
 import './frontend/index.css';
 import './frontend/theme.css';
-import { renderComponent } from './app/bridge/componentRenderer';
+import { renderComponent } from './app/webserver/componentRenderer';
 
 // Parse URL parameters to determine which component to render
 const params = new URLSearchParams(window.location.search);
--- a/src/main.ts
+++ b/src/main.ts
@@ -5,7 +5,7 @@ import { setupTaskbar } from './app';
 import { publishDashboardUpdates, dashboardBridge } from './app/bridge/dashboard/dashboardBridge';
 import { TelemetrySink } from './app/bridge/iracingSdk/telemetrySink';
 import { OverlayManager } from './app/overlayManager';
-import { startComponentServer } from './app/bridge/componentServer';
+import { startComponentServer } from './app/webserver/componentServer';
 import { updateElectronApp } from 'update-electron-app';
 // @ts-expect-error no types for squirrel
 import started from 'electron-squirrel-startup';


src/app/bridge/bridgeProxy.ts
```diff
@@ -0,0 +1,160 @@
...
+    bridge.onRunningState((running: boolean) => {
+      isRunning = running;
+      console.log('🔄 Bridge proxy: Received running state:', running, ', broadcasting to', io.engine.clientsCount, 'clients');
+      io.emit('runningState', running);
+    });
+    console.log('✅ Bridge proxy: onRunningState listener registered');
+  };
+
+  // Subscribe to initial bridge
+  subscribeToBridge(irsdkBridge);
```
tariknz: Just wondering, do you think it would be good to only enable the listeners if there's a socket client actually connected. e.g. 
```
// Only subscribe when first client connects
io.on('connection', (socket) => {
  if (io.engine.clientsCount === 1) {
    subscribeToBridge(irsdkBridge); // First client
  }
  
  socket.on('disconnect', () => {
    if (io.engine.clientsCount === 0) {
      unsubscribeFromBridge(); // Last client disconnected
    }
  });
});
```

src/app/bridge/bridgeProxy.ts
```diff
@@ -0,0 +1,160 @@
...
+  console.log('🔍 Bridge has onSessionData?', typeof irsdkBridge.onSessionData);
+  console.log('🔍 Bridge has onRunningState?', typeof irsdkBridge.onRunningState);
+
+  // Function to subscribe to bridge events
+  const subscribeToBridge = (bridge: IrSdkBridge) => {
+    console.log('🔌 Bridge proxy: Subscribing to bridge events...');
+    
+    bridge.onTelemetry((telemetry: Telemetry) => {
+      currentTelemetry = telemetry;
+      console.log('🔄 Bridge proxy: Received telemetry, broadcasting to', io.engine.clientsCount, 'clients');
```
tariknz: These logs are you planning to keep them, they will be pretty busy. Specifically the onTelemetry and onSessionData ones. You could sample them if you prefer.

src/app/webserver/bridgeProxy.ts
```diff
@@ -0,0 +1,160 @@
...
+      io.emit('runningState', running);
+    });
+    console.log('✅ Bridge proxy: onRunningState listener registered');
+  };
+
+  // Subscribe to initial bridge
+  subscribeToBridge(irsdkBridge);
+
+  // Return a function to resubscribe when bridge changes
+  const resubscribeToBridge = (newBridge: IrSdkBridge) => {
```
tariknz, line 71: what would be the scenario that causes resubscribeToBridge to be called? The main issue here is I think you might have a memory leak where subscribeToBridge is setting up event listeners but they never clear so if this runs again it will create duplicate listners.

tariknz: I can see actually this happens at least when you toggle demo mode

src/app/bridge/componentServer.ts
```diff
@@ -0,0 +1,250 @@
...
+      components: componentNames,
+      baseUrl: `http://localhost:${COMPONENT_PORT}`,
+      websocketUrl: `ws://localhost:${COMPONENT_PORT}`,
+      examples: componentNames.map(
+        (name) => `http://localhost:${COMPONENT_PORT}/component/${name}`
+      ),
+    });
+  });
+
+  httpServer.listen(COMPONENT_PORT, () => {
```
tariknz: Do you want to just bind to localhost? 

```suggestion
  httpServer.listen(COMPONENT_PORT, 'localhost', () => {
```

This avoids risking having them available via the network or possibly the internet. This might be desirable but we should probably make it a setting.

src/frontend/components/Standings/Standings.tsx
```diff
@@ -38,6 +38,24 @@ export const Standings = () => {
     return <></>;
   }
 
+  // If no data, show a message instead of empty table
```
tariknz: Are we able to get rid of the "Waiting for session data..." UI because they are going to appear for normal widgets. We could condditionally make them appear just for the web views if they are necessary. My concern is they probably shouldn't live per component though we can probably do something better.

src/frontend/components/Input/Input.tsx
```diff
@@ -1,10 +1,27 @@
...
+import { InputWidgetSettings } from '../Settings/types';
 
-export const Input = () => {
+export const Input = (props?: InputWidgetSettings['config']) => {
   const inputs = useInputs();
-  const settings = useInputSettings();
+  const settingsFromHook = useInputSettings();
+
+  // Use props if provided (browser mode), otherwise use hook (Electron mode)
+  const settings = props || settingsFromHook;
```
tariknz: Why was this necessary?

src/frontend/components/Input/InputTrace/InputTrace.tsx
```diff
@@ -89,6 +89,7 @@ export const InputTrace = ({
     <svg
       ref={svgRef}
       width={'100%'}
+      height={'100%'}
```
tariknz: What was the issue that needed you to update the styles for Input?
Do you need help with that? Happy to take a look. The input specifically was a little tricky to try to get it to be responsive so you're probably battling with that i suspect.

tariknz: Currently the bars look pretty chonky and there's a gap to the right of the wheel:

Image

src/app/bridge/componentWindow.ts
```diff
@@ -0,0 +1,187 @@
...
+  <body>
+    <div id="app" class="loading">Loading ${componentName}...</div>
+    <script type="module">
+      import { createRoot } from 'https://esm.sh/react-dom@19';
+      import {
+        TelemetryProvider,
+        DashboardProvider,
+        RunningStateProvider,
+        SessionProvider,
+      } from 'https://esm.sh/@irdashies/context';
```
tariknz: Is there a way to reference the package locally instead? Like the below WidgetIndex? 
I'm not familiar with esm.sh but I suspect this will get you in a world of pain since it would mean the released version would be out of sync with I assume the hosted one would be master.

src/frontend/context/SessionStore/SessionProvider.tsx
```diff
@@ -16,13 +16,13 @@ export const SessionProvider = ({ bridge }: SessionProviderProps) => {
...
       });
-      return () => bridge.then((bridge) => bridge.stop());
+      return;
     }
 
     bridge.onSessionData((telemetry) => {
       setSession(telemetry);
     });
-    return () => bridge.stop();
+    // Don't stop bridge on unmount - it should persist
```
tariknz: How come? What problem is being solved here?

src/frontend/theme.css
```diff
@@ -8,6 +8,7 @@ layer(base);
   --text-xs: 10px;
   --text-sm: 12px;
   --text-base: 14px;
+  --spacing: 0.25rem;
```
tariknz: What's this for?

src/app/bridge/bridgeProxy.ts
```diff
@@ -0,0 +1,160 @@
+import { Server as HTTPServer } from 'http';
+import { Server as SocketIOServer, Socket } from 'socket.io';
```
tariknz: One for later, we should've really considered using ws instead of socket.io
From reading it seems like socket.io has a much higher overhead, latency, and lower badwidth compared to ws

This is farily straight forward: https://www.npmjs.com/package/ws https://github.com/websockets/ws

package.json
```diff
@@ -87,11 +88,13 @@
     "@phosphor-icons/react": "^2.1.10",
     "d3": "^7.9.0",
     "electron-squirrel-startup": "^1.0.1",
+    "express": "^5.1.0",
```
tariknz: Is express necessary? We probably could just use http there's no need for the additional depednency for pretty basic usage here.

src/frontend/components/Input/InputBar/InputBar.tsx
```diff
@@ -34,84 +34,92 @@ export const InputBar = ({
...
-    .attr('class', 'value')
-    .attr('x', (_, i) => (xScale(i.toString()) ?? 0) + xScale.bandwidth() / 2)
-    .attr('y', () => 10)
-    .attr('text-anchor', 'middle')
-    .attr('font-size', '10px')
-    .attr('fill', 'white')
-    .text((d) => (d.value * 100).toFixed(0));
-
-  // Overlay ABS label if needed
-  if (brakeAbsActive && includeAbs) {
```
tariknz, line 100: Did you mean to get rid of the ABS functionality?

PR #106, opened 2025-11-18
feat: session title bar

LanceVShoot:
Thoughts on this?

I created a title bar in simhub that has a background that fills depending on the completion percentage of the race. So I added that into a title bar. The title bar displays the name of the session, which I don't know if we can/should include it. Just because it doesn't come from the sdk (as far as I can tell it's not in there anywhere). So I had to create a static mapping of the series IDs with their names - that would need to be updated.

I also divided out the "Show only when on track" option to allow independent control of standings/relatives.

Screenshot 2025-11-18 093845

### DIFF
--- a/src/frontend/components/Standings/Standings.tsx
+++ b/src/frontend/components/Standings/Standings.tsx
@@ -7,7 +7,6 @@ import { RatingChange } from './components/RatingChange/RatingChange';
 import { SessionBar } from './components/SessionBar/SessionBar';
 import { SessionFooter } from './components/SessionFooter/SessionFooter';
 import { TitleBar } from './components/TitleBar/TitleBar';
-import { useDashboard } from '@irdashies/context';
 import {
   useCarClassStats,
   useDriverStandings,
@@ -33,7 +32,6 @@ export const Standings = () => {
   const classStats = useCarClassStats();
   const isMultiClass = standings.length > 1;
   const highlightColor = useHighlightColor();
-  const { currentDashboard } = useDashboard();
 
   // Show only when on track setting
   if (settings?.showOnlyWhenOnTrack && !isDriving) {


src/app/irsdk/node/irsdk-node.ts
```diff
@@ -165,8 +165,30 @@ export class IRacingSDK {
 
     try {
       const seshString = this._sdk?.getSessionData();
-      // Remove trailing commas
-      const fixedYaml = seshString?.replace(/(\w+):\s*,\s*\n/g, '$1: \n');
+      // Fix YAML issues: quote string values that contain commas, treat empty/whitespace/comma values as null
```
tariknz: What's this trying to fix here? 
Just one thing to consider is this code here runs at 25hz from memory so its a pretty hot path and there's a bit of logic going on here.

LanceVShoot: It's only purpose is to get irdashies to work in offline/ai races. Without it, the yaml has parsing issues and fails. So it could likely be handled a different way, maybe with how the player data is returned to enforce certain values if null, etc?

tariknz: That's strange AI races and test drive should work. This must've been a recent change that broke it from iRacing. I'll take a closer look.

LanceVShoot: This change definitely breaks the live standings/relatives. Abort, abort!

LanceVShoot:
just cleaning this draft pr up a bit after I incorrectly modified the irsdk ts file

LanceVShoot:
The color is the same as your highlight color you have selected! ;)

LanceVShoot:
Also 100% fine defaulting to disabled. Anything I add is most likely selfishly for myself. So Incan always enable it if needed! :)

tariknz:
Yep let's do default disabled then roll it out
That's ok that was part of my thinking for this project, anyone that wants a feature can choose to build it we just keep everything behind settings :)

Its probably going to get messy at some point but we can cross that bridge when we need to to do a reorg of all the settings.

PR #109, opened 2025-11-20
feat: add gap and interval

LanceVShoot:
in standings, there's currently a delta setting which displays delta to the overall leader (even in multi class, delta is to overall leader and not class leader). Added gap and interval that will probably replace delta? And adjusted so gap is relative to class leader instead of overall leader.


LanceVShoot:
Updated to use just delta for calculations instead of adding new calculations, etc.

LanceVShoot:
This is ready!

PR #111, opened 2025-11-21
Fix minor formatting of manufacturer/tire/flag

LanceVShoot:
The manufacturer/tire/flag icons were being set to defined sizes. Instead they should just fill their container/cell. Removed the sizing component from them and updated stories.


LanceVShoot:
Font size medium (old)
medium-old

Font size medium (this update)
medium-new

Font size 3XL (old)
3xl-old

Font size 3XL (this update)
3xl-new

PR #120, opened 2025-11-22
Session bar time format

LanceVShoot:
Session bar time was showing in minutes only. Added formatting.

Before
1282:41 / 1440 m
30:12 / 35 m

After
21:22:41 / 24 hr
30:12 / 35 min


tariknz:
Is this PR still relevant with the other one doing the format settings? Or can we close it?

PR #125, opened 2025-11-24
Feat/fuel calculator enhancements 1

23Kev:
### New features

- Consumption history charts, histogram to show 30 laps history, line to show 5 laps
- Endurance Strategy section that shows how many stops will need to be completed and the calculated stint length
- Horizontal layout option. Can be used along the top of the screen or line it up with another overlays. 
- Enable/disable components to reduce the size to align it with other overlays.

### Enhancements

- Laps remaining now shows partial laps to help to know if you could make another full lap
- More reliable calculation of first lap and last lap
- Colour coded all values and overlay border. Green - safe on fuel, orange - close to needing to pit, red - must pit to finish the race
- More tabular like alignment for the Min/Max/Last and Avg fuel consumption. This section now includes values to show how much fuel is needed at each consumption level to finish the race/stint
- Safety Margin is now used in all calculations.
- Fuel consumption figures will reset at the end of the race session, but will carry through from Warmup > Qualifying > Race
- Improved performance, memory and CPU usage by ensuring we only calculate when things change, not at every update from the sim. This was a full refactor of all files used in the Fuel Calculator and is a substantial change.
- Extensive logging included in all parts of the calculator, update const DEBUG_LOGGING = true; (in useFuelCalculator) to enable it.


tariknz:
Hey @23Kev is this ready?

23Kev:
> Hey @23Kev is this ready?

I still need to do some more testing, I want to make sure this is as good as it can be taking into account all the feedback I've received. I'll need another couple of days I'd say.

PR #126, opened 2025-11-24
Time format options

LanceVShoot:
Updated session bar to use new FormatTime() so that elapsed time and duration of race show correctly. Previously they would only show in seconds. Also updated the session bar to appropriately determine whether to show the time remaining or not. Previously, if a session was less than or equal to 24hrs, it would assume the session is time-based (not lap-based). That's because iRacing sets lap-based sessions to 24h10m or something. However if a session is created that's 24h30m long, it didn't show the time because it was greater than 24hrs. Instead, we now look at sessionLaps to determine if it's lap or time-based. sessionLaps is 'unlimited' when the session is time-based.


src/frontend/components/Standings/components/SessionBar/SessionBar.tsx
LanceVShoot, line 1: fyi also adjusted session bar flex so inner containers expand/grow naturally based on their contents, instead of all containers being split evenly by size. Previously if a container had more data that the split width, text wrapped. (23:22:05 / 24 Hrs for example). And really the containers should have overflow hidden probably will look at that later. This new way should be beneficial when we separate out the session bars and allow components to be customized/dropped in by the user.

PR #130, opened 2025-11-25
Updated relative and standings to focus on the currently spectated driver

23Kev:
Removed filter to only update when the isSpectator flag is true.
Now any change to the camera car should update the relative and standings.


src/frontend/context/shared/useFocusCarIdx.ts
tariknz:
```suggestion
import { useDriverCarIdx } from '../SessionStore/SessionStore';
```

PR #131, opened 2025-11-26
feat: update tire compound to svg

LanceVShoot:
using svg instead of image for better visibility. Also updated compound to show correctly based on the tire the driver currently has selected. (different cars have different compounds/names, so some cars weren't showing correctly)


LanceVShoot:
Looking for any guidance on how to map car tire compounds to different cars. For example, NASCAR uses yellow for their "Primary". But other cars (IndyCar) uses white for their "Primary". Is there a mapping somewhere that we can pull perhaps? Instead of manually trying to figure all of these out?

Screenshot 2025-11-25 171729
Screenshot 2025-11-25 171757
Screenshot 2025-11-25 171818
Screenshot 2025-11-25 173816

tariknz (approved):

PR #132, opened 2025-11-26
show laps down for gap

LanceVShoot:
show laps down for gap instead of showing seconds when driver is not on same lap as cars ahead

### DIFF
--- a/src/frontend/components/Standings/createStandings.ts
+++ b/src/frontend/components/Standings/createStandings.ts
@@ -329,7 +329,8 @@ export const augmentStandingsWithGap = (
       // Calculate laps down: positive if driver is behind on laps, negative if ahead
       const leaderLap = lastLap[classLeader.carIdx] ?? 0;
       const driverLap = lastLap[driverStanding.carIdx] ?? 0;
-      const lapsDown = Math.max(0, leaderLap - driverLap); // Only show positive laps down
+      // Calculate laps down: only show when leader is at least 2 laps ahead
+      const lapsDown = Math.max(0, leaderLap - driverLap - 1);
 
       // Gap is simply the difference between this driver's delta and the class leader's delta
       const gap = driverStanding.delta !== undefined && classLeader.delta !== undefined


src/frontend/components/Standings/createStandings.ts
```diff
@@ -314,16 +313,21 @@ export const augmentStandingsWithGap = (
...
       if (driverStanding.carIdx === classLeader.carIdx) {
         // Class leader shows as dash (undefined gap)
-        return { ...driverStanding, gap: undefined };
+        return { ...driverStanding, gap: undefined, lapsDown: 0 };
       }
 
+      // Calculate laps down: positive if driver is behind on laps, negative if ahead
+      const leaderLap = lastLap[classLeader.carIdx] ?? 0;
+      const driverLap = lastLap[driverStanding.carIdx] ?? 0;
+      const lapsDown = Math.max(0, leaderLap - driverLap); // Only show positive laps down
```
tariknz: Just thinking when the leader passes the start/finish line dos that mean everyone else would show as a lap down until they start crossing? Maybe we should only make it appear when > 1 lap

LanceVShoot: Good catch! Updated so laps down aren't shown until driver ahead is on a lap that's at least 2 ahead of the driver behind. Otherwise time behind is shown.

tariknz: iRacing also only shows 2+ laps down

src/frontend/components/Standings/createStandings.ts
```diff
@@ -324,16 +323,22 @@ export const augmentStandingsWithGap = (
...
         // Class leader shows as dash (undefined gap)
-        return { ...driverStanding, gap: undefined };
+        return { ...driverStanding, gap: undefined, lapsDown: 0 };
       }
 
+      // Calculate laps down: positive if driver is behind on laps, negative if ahead
+      const leaderLap = lastLap[classLeader.carIdx] ?? 0;
+      const driverLap = lastLap[driverStanding.carIdx] ?? 0;
+      // Calculate laps down: only show when leader is at least 2 laps ahead
+      const lapsDown = Math.max(0, leaderLap - driverLap - 1);
```
tariknz, line 333: One other thing I just remembered - you probably want to do this just for race sessions. I think everyone will be on different laps for practice and quali its going to be very confusing.

PR #161, opened 2025-12-21
feat: blind spot monitor

tariknz:
Adds blind spot monitor widget:

blindspot-monitor

- Left and right side detection
- Visual amber indicator bars that move vertically based on car distance
- Detection of single car or multiple cars (2 cars) on each side
- Configurable detection distances:
  - Distance ahead (3-6 meters)
  - Distance behind (3-6 meters)
- Adjustable indicator width (5-100px)
- Adjustable background opacity
- Only displays when on track
- Real-time position tracking based on lap distance

https://github.com/user-attachments/assets/75862879-135d-4e23-bea5-7b3443377251

Screenshot 2025-12-21 at 1 56 55 PM

### DIFF
--- a/src/app/storage/defaultDashboard.ts
+++ b/src/app/storage/defaultDashboard.ts
@@ -388,7 +388,7 @@ export const defaultDashboard: DashboardLayout = {
     },
     {
       "id": "blindspotmonitor",
-      "enabled": false,
+      "enabled": true,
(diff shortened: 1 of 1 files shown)


FSund:
This was the final feature i felt irdashies was missing. It now has everything I could ever want. Fantastic work!

PR #191, opened 2025-12-30
feat: Introducing Precipitation (%) as a property

RenatoCoicev:
Introducing Precipitation (%) as a property to be displayed on footer and header of both Relatives and Standings.

I had to regenerate `src/app/irsdk/types/_GENERATED_telemetry.ts` because it was missing the Precipitation prop. A bunch of other props were added or ended up in a different line, but the file is as generated, no manual changes.

### DIFF
--- a/src/app/storage/defaultDashboard.ts
+++ b/src/app/storage/defaultDashboard.ts
@@ -400,7 +400,7 @@ export const defaultDashboard: DashboardLayout = {
         "showMax": true,
         "showPitWindow": true,
         "showEnduranceStrategy": false,
-        "showFuelSave": true,
+        "showFuelScenarios": true,
         "showFuelRequired": false,
         "showConsumptionGraph": true,
         "consumptionGraphType": "histogram",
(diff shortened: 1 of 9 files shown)


tariknz (approved):
Nice work no comments here!
LGTM

tariknz (changes requested):
Just reiterating what I wrote in discord after seeing the failing lint.  So **dcPeakBrakeBias** only appears when you're driving the Clio so we might need to manually always include that TelemetryVar. Right now its failing the linting but other than that everything looks great!

RenatoCoicev:
> Just reiterating what I wrote in discord after seeing the failing lint. So **dcPeakBrakeBias** only appears when you're driving the Clio so we might need to manually always include that TelemetryVar. Right now its failing the linting but other than that everything looks great!

Yeah, it seems like `irdashies\src\app\irsdk\types\_GENERATED_telemetry.ts` is generated based in the session that is open at the moment in iRacing. Not good. I loaded a PCC practice sessions but still got no `dcPeakBrakeBias`, probably because I loaded in the GR86 as I don't own the Clio. I manually forced `irdashies\src\app\irsdk\native\scripts\generate-var-types.js` to include dcPeakBrakeBias for now and manually checked so we are not missing any other prop that was present in previously versions.

To make this future proof, the recreate process would need to be adding new properties only, not removing. But that could leave us with deprecated properties. #techDebt

tariknz (approved):

PR #206, opened 2026-01-01
Add session-specific visiblity options

Grollo13:
This PR adds session specific visibility options to all Widgets.

This allows the user to enable Widgets depending on the current session. Some driver might want to disable the standings during qualifying to avoid distractions, other might want to see the laptimes of others. 

grafik

### DIFF
--- a/src/frontend/components/Settings/types.ts
+++ b/src/frontend/components/Settings/types.ts
@@ -256,3 +256,9 @@ export interface GarageCoverWidgetSettings extends BaseWidgetSettings {
     imageFilename: string;
   };
 }
+export interface FasterCarsFromBehindWidgetSettings extends BaseWidgetSettings {
+  config: {
+    distanceThreshold: number;
+    sessionVisibility: SessionVisibilitySettings;
+  };
+}
--- a/src/app/storage/defaultDashboard.ts
+++ b/src/app/storage/defaultDashboard.ts
@@ -404,7 +404,8 @@ export const defaultDashboard: DashboardLayout = {
         "height": 43
       },
       "config": {
-        "distanceThreshold": -0.3
+        "distanceThreshold": -0.3,
+        "sessionVisibility": { race: true, loneQualify: true, openQualify: true, practice: true, offlineTesting: true }
       }
     },
     {
--- a/src/frontend/components/FasterCarsFromBehind/FasterCarsFromBehind.tsx
+++ b/src/frontend/components/FasterCarsFromBehind/FasterCarsFromBehind.tsx
@@ -17,9 +17,13 @@ export const FasterCarsFromBehind = () => {
     distanceThreshold: settings?.distanceThreshold,
   });
 
-  if (sessionType === 'Lone Qualify') {
-    return <></>;
-  }
+  
+  // Show only in selected sessions
+  if (sessionType === 'Race' && !settings?.sessionVisibility.race) return <></>;
+  if (sessionType === 'Lone Qualify' && !settings?.sessionVisibility.loneQualify) return <></>;
+  if (sessionType === 'Open Qualify' && !settings?.sessionVisibility.openQualify) return <></>;
+  if (sessionType === 'Practice' && !settings?.sessionVisibility.practice) return <></>;
+  if (sessionType === 'Offline Testing' && !settings?.sessionVisibility.offlineTesting) return <></>;
 
   return <FasterCarsFromBehindDisplay {...carBehind} />;
 };
--- a/src/frontend/components/FasterCarsFromBehind/hooks/useFasterCarsSettings.tsx
+++ b/src/frontend/components/FasterCarsFromBehind/hooks/useFasterCarsSettings.tsx
@@ -1,11 +1,5 @@
 import { useDashboard } from '@irdashies/context';
-
-interface FasterCarsFromBehindSettings {
-  enabled: boolean;
-  config: {
-    distanceThreshold: number;
-  };
-}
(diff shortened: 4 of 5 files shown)


src/frontend/components/Standings/Standings.tsx
```diff
@@ -29,6 +30,14 @@ export const Standings = () => {
...
+
+  // Show only in selected sessions
+  if (sessionType === 'Race' && !settings?.sessions.race) return <></>;
+  if (sessionType === 'Lone Qualify' && !settings?.sessions.loneQualify) return <></>;
+  if (sessionType === 'Open Qualify' && !settings?.sessions.openQualify) return <></>;
+  if (sessionType === 'Practice' && !settings?.sessions.practice) return <></>;
+  if (sessionType === 'Offline Testing' && !settings?.sessions.offlineTesting) return <></>;
 
   // Show only when on track setting
   if (settings?.showOnlyWhenOnTrack && !isDriving) {
```
tariknz, line 48: I think this is good pretty straight forward but if we are to reuse this logic we'll probably just want to compose the component instead.

e.g. we'd end up with something like this:
```
<DisplayGuard
  displayMode={{
    sessions: settings?.sessions,
    showOnlyWhenOnTrack: settings?.showOnlyWhenOnTrack,
  }}
>
  {/* widget content */}
</DisplayGuard>
```

src/frontend/components/Settings/types.ts
```diff
@@ -58,6 +58,13 @@ export interface StandingsWidgetSettings extends BaseWidgetSettings {
     driverName: { enabled: boolean };
     pitStatus: { enabled: boolean };
     displayOrder: string[];
+    sessions: {
+      race: boolean;
+      loneQualify: boolean;
+      openQualify: boolean;
+      practice: boolean;
+      offlineTesting: boolean;
+    }
```
tariknz: You would want to assume this optional as this won't exist in the dashboard.json for existing users.

src/frontend/components/Standings/Standings.tsx
```diff
@@ -29,6 +30,14 @@ export const Standings = () => {
   const numCarClasses = useWeekendInfoNumCarClasses();
   const isMultiClass = (numCarClasses ?? 0) > 1;
   const highlightColor = useHighlightColor();
+  const sessionType = useCurrentSessionType();
+
+  // Show only in selected sessions
+  if (sessionType === 'Race' && !settings?.sessions.race) return <></>;
+  if (sessionType === 'Lone Qualify' && !settings?.sessions.loneQualify) return <></>;
```
tariknz: Since this is optional you should shortcircuit this so if the sessions object doesn't exist then assume we're showing for every session. For backward compatibility. Else currently these would fail (session object doesn't exist) but even if we make sessions optional it would hide for every case since the value of the properties will always be falsy if they don't exist.

src/frontend/components/Settings/sections/StandingsSettings.tsx
```diff
@@ -765,6 +773,92 @@ export const StandingsSettings = () => {
                 }
               />
             </div>
+
+            {/* Session Settings */}
```
tariknz: This settings section might also need to be abstracted so its reusable for other components. But again we can do the bastractions in a seperate PR if you want to keep this clean as a first version.

src/frontend/components/Settings/types.ts
Grollo13: Isn't this resolved by `migrateConfig`?

I tested it like this:
- Run irDashies 0.0.35+5e76d8f
- Go to Advanced and reset configuration
- Confirm that session settings are not included in the json anymore
- Close irDashies
- Run irDashies from this branch
-> Default configuration is present for new settings
-> No errors/warnings observed

src/frontend/components/Standings/Standings.tsx
Grollo13: I've been looking into this now:

Ideally, the user of this component should not be concerned about the internals, e.g. which conditions are about to be checked. So I'd like to replace `sessions`/`showOnlyWhenOnTrack` with the widget settings instead.

But this would require a bit of refactoring of the settings types. The session/onTrack visibility settings would need to be moved into `BaseWidgetSettings`. But the `use<Widget>Settings` currently return not the settings object iself, but the `config` property instead. So this would need to be changed for all occurrences.

I think for now it would be a better option to use the simple conditions introduced in this PR, and introduce the `DisplayGuard` component in a separate PR.

src/frontend/components/Settings/sections/StandingsSettings.tsx
Grollo13: Moved it into `src\frontend\components\Settings\components\SessionVisibility.tsx`

src/frontend/components/Standings/Standings.tsx
tariknz: Yep fair enough!

tariknz:
> > Happy with this, feel free to publish if you're done and we can get it approved and merged.
> 
> @tariknz Alright, I started to update the other components. But there are some differences in terms of how default settings are defined for example. Which approach is the preferred one? Should the default config be in `<Widget>Settings.tsx` or `useWidgetSettings`? I prefer the last one.

Yeah that's probably better to use the latter so it makes sure the widget using it has its defaults set correctly. Its a bit of a mess right now. I want to eventually have the source of truth to be the `defaultDashboard.ts` for defaults instead of each setting trying to define its defaults and fallback.

PR #207, opened 2026-01-02
feat: add 'Use Live Position Standings' feature to Standings and Relatives overlays

RenatoCoicev:
Adding a new setting to Standings and Relatives overlays to use live car position. When set to true, overlay will use live car position that is constantly updated rather than the iRacing car position that updates only when a car crosses the start/finish line.


src/frontend/components/Standings/hooks/useDriverLivePositions.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * Live position is calculated using CarIdxLapCompleted plus CarIdxLapDistPct
+ * from telemetry, then sorted to return integer positions starting from 1.
+ * Positions are relative to the driver's class.
+ *
+ * @returns Record<driverId, position> where position is an integer relative to the driver's class
+ */
+export const useDriverLivePositions = (): Record<number, number> => {
+  const carIdxLapCompleted = useTelemetryValues<number[]>('CarIdxLapCompleted');
+  const carIdxLapDistPct = useTelemetryValues<number[]>('CarIdxLapDistPct');
+  const CarIdxClass = useTelemetryValues<number[]>('CarIdxClass');
```
tariknz: Minor but this should be lowercase 


```suggestion
  const carIdxClass = useTelemetryValues<number[]>('CarIdxClass');
```

src/frontend/components/Standings/hooks/useDriverPositions.tsx
```diff
@@ -115,6 +117,9 @@ export const useCarState = () => {
 // currently there's still a few bugs to handle but is only used in relative right now
 export const useDriverStandings = () => {
   const driverPositions = useDriverPositions();
+  const driverLivePositions = useDriverLivePositions();
+  const relativeSettings = useRelativeSettings();
```
tariknz, line 121: This should be in standingsSettings right? Why relative?

tariknz: Ah I think i see what's going on this is my mess with the hooks.

src/frontend/components/Standings/hooks/useDriverLivePositions.tsx
RenatoCoicev: Sorted, thanks!

src/frontend/components/Standings/hooks/useDriverPositions.tsx
RenatoCoicev: Both useDriverPosition and useDriverStandings export a useDriverStandings. But I made them both consume from the new hook useDriverLivePosition so it shouldn't be that terrible in the future when we combine them.

The comment says
// TODO: this should eventually replace the useDriverStandings hook
// currently there's still a few bugs to handle but is only used in relative right now

PR #210, opened 2026-01-03
feat: hide car logo if single make session

Xenon1:
### Overview
New feature to optionally hide car logo's if there is only 1 car make in the session, This is ignored for multiclass

Added config option to Relative and Standings that is only visable if **Car Manufacturer** is enabled
image

### Teating
- F4 with **Car Manufacturer** enabled but **Hide If Single Make** disabled 
image

- F4 with new config enabled
image

- PCC with new config enabled (Ensure multiclass exception is working, even with 4 classes of the same cars)
image

- GT3 with new config enabled (Ensure cat logo's still show)
image

_Excuse the messy commit im not normally this bad_


.vs/ProjectSettings.json
tariknz, line 1: Can you remove the .vs folder 
Since its ignored you need to do:
`git rm -r --cached .`
Then commit / push

src/frontend/components/Standings/hooks/useIsSingleMake.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Returns true when the session contains only a single car manufacturer.
+ * If the session is multi-class (NumCarClasses > 1) this will always return false
+ * so manufacturers remain visible in multi-class sessions.
+ */
+export const useIsSingleMake = () => {
+  const sessionDrivers = useSessionDrivers();
+  const numCarClasses = useWeekendInfoNumCarClasses();
+
+  return useMemo(() => {
+    if (!sessionDrivers || sessionDrivers.length === 0) return false;
```
tariknz: Minor, but this is the same:
```suggestion
    if (!sessionDrivers?.length) return false;
```

src/frontend/components/Standings/hooks/useIsSingleMake.ts
```diff
@@ -0,0 +1,32 @@
...
+        CAR_ID_TO_CAR_MANUFACTURER[driver.CarID]?.manufacturer ?? 'unknown';
+      manufacturers.add(manufacturer);
+      if (manufacturers.size > 1) return false;
+    }
+
+    return manufacturers.size === 1;
+  }, [sessionDrivers, numCarClasses]);
+};
+
+export default useIsSingleMake;
```
tariknz:
```suggestion
```
No need to export default, you're already exporting it. Its better to just export what you want like the above.

tariknz (approved):
lgtm

PR #211, opened 2026-01-03
feat: implement global toggle for UI overlays

RenatoCoicev:
This is a initial implementation of a "hide overlay" mechanism. It's a global hide-all-overlays hardcoded to Alt+H. In the future I'd like to make this configurable per overlay, allowing user to set shortcuts to hide each overlay.

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+// App.tsx
 import { createRoot } from 'react-dom/client';
 import { HashRouter, Route, Routes } from 'react-router-dom';
 import {
@@ -13,6 +13,7 @@ import { Settings } from './components/Settings/Settings';
 import { EditMode } from './components/EditMode/EditMode';
 import { ThemeManager } from './components/ThemeManager/ThemeManager';
 import { WIDGET_MAP } from './WidgetIndex';
+import { HideUIWrapper } from './components/HideUIWrapper';
 
 const AppRoutes = () => {
   const { currentDashboard } = useDashboard();
@@ -30,13 +31,7 @@ const AppRoutes = () => {
           <Route
             key={widget.id}
             path={`/${widget.id}`}
-            element={
-              running ? (
-                <WidgetComponent {...widget.config} />
-              ) : (
-                <></>
-              )
-            }
+            element={running ? <WidgetComponent {...widget.config} /> : <></>}
           />
         );
       })}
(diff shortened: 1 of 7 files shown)


src/frontend/App.tsx
```diff
@@ -44,26 +45,49 @@ const AppRoutes = () => {
...
+
+const App = () => {
+  const [hideUI, setHideUI] = useState(false);
+
+  useEffect(() => {
+    if (window.globalKey?.onToggle) {
+      const unsub = window.globalKey.onToggle((hide) => setHideUI(hide));
+      return () => unsub();
+    }
+  }, []);
```
tariknz: Can you move this to its own component?
It will be a lot cleaner than adding business logic in the main App.tsx

src/main.ts
```diff
@@ -41,5 +41,31 @@ app.on('ready', async () => {
...
+  if (!registered) {
+    console.error(`Failed to register global shortcut: ${accel}`);
+  }
+});
+
+app.on('will-quit', () => {
+  globalShortcut.unregister('Alt+H');
+  globalShortcut.unregisterAll();
+});
+
```
tariknz: Please move this logic to its own file also  - I want to keep the main.ts entry point really lean so its easy to follow.

src/frontend/App.tsx
RenatoCoicev: Sure thing, thanks for the heads up! I'll let you know once I have it ready again, cheers.

src/main.ts
```diff
@@ -41,5 +41,31 @@ app.on('ready', async () => {
...
+    // Toggle internal state and notify renderer
+    hideState = !hideState;
+    overlayManager.getOverlays().forEach(({ window }) => {
+    window.webContents.send('global-toggle-hide', hideState);
+  });
+  });
+
+  if (!registered) {
+    console.error(`Failed to register global shortcut: ${accel}`);
+  }
```
tariknz: This logic can go into app.on('ready') but make sure its in its own class or function.

src/preload.ts
```diff
@@ -1,5 +1,14 @@
...
 
 exposeBridge();
+
+contextBridge.exposeInMainWorld('globalKey', {
+  onToggle: (cb: (hide: boolean) => void) => {
+    const listener = (_: Electron.IpcRendererEvent, hide: boolean) => cb(hide);
+    ipcRenderer.on('global-toggle-hide', listener);
+    return () => ipcRenderer.removeListener('global-toggle-hide', listener);
+  }
+});
```
tariknz: Please also create a rendererExpose function in a seperate file similar to the expose bridge function.

src/frontend/App.tsx
```diff
@@ -44,26 +45,49 @@ const AppRoutes = () => {
...
+
+  return (
+    <div className={hideUI ? 'hidden-ui' : ''}>
```
tariknz:
```suggestion
    <div className={hideUI ? 'opacity-0 pointer-events-none transition-opacity duration-100 ease-linear
' : ''}>
```
Don't create CSS classes - use tailwind please

src/frontend/index.css
```diff
@@ -8,3 +8,9 @@ html,body,#app {
   width: 100%;
   height: 100%;
 }
+
+.hidden-ui {
+  opacity: 0;
+  pointer-events: none;
+  transition: opacity 120ms linear;
+}
```
tariknz: This can go once converting to tailwind classes

PR #215, opened 2026-01-04
feat: Add LapCounter to Header/Footer bars

Grollo13:
Includes a total lap time estimation for timed races

### DIFF
--- a/src/frontend/context/shared/useTotalRaceLaps.tsx
+++ b/src/frontend/context/shared/useTotalRaceLaps.tsx
@@ -1,0 +1,75 @@
+import { useSessionLapCount } from '../../components/Standings/hooks/useSessionLapCount';
+import {
+  useCurrentSessionType,
+  useFocusCarIdx,
+  useTelemetryValue,
+  useCarIdxClassEstLapTime,
+  useTelemetryValues
+} from '@irdashies/context';
+
+// Estimate the total number of laps that will be completed by the drivers car in a timed session.
+export function useTotalRaceLaps(): number | null {
+  const carIdx = useFocusCarIdx() as number;
+  const { timeRemaining, timeTotal, totalLaps } = useSessionLapCount();
+  const lap = useTelemetryValue("Lap") as number;
+  const sessionType = useCurrentSessionType();
+  const lapDistPct = useTelemetryValue('LapDistPct');
+  const carIdxLap = useTelemetryValues('CarIdxLap');
+  const carIdxPosition = useTelemetryValues('CarIdxPosition');
+  const carIdxLapDistPct = useTelemetryValues('CarIdxLapDistPct');
+  // TODO: Switch to useCarIdxAverageLapTime()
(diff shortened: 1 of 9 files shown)


src/frontend/components/FuelCalculator/useFuelCalculation.tsx
```diff
@@ -338,18 +339,13 @@ export function useFuelCalculation(
...
-        const estLapTime =
-          validLaps.reduce((sum, l) => sum + l.lapTime, 0) / validLaps.length;
-        lapsRemaining = Math.ceil(sessionTimeRemain / estLapTime) + 1;
-        totalLaps = lap + lapsRemaining;
-      } else {
-        // Can't determine laps remaining for timed race - use best estimate
-        // Assume 1 lap remaining to keep calculator visible
-        lapsRemaining = 1;
-        totalLaps = lap + 1;
+      const totalRaceLaps = useTotalRaceLaps();
```
tariknz: Yeah I'm pretty sure you can't use hooks conditionally like that. 

If the hook is shared across multiple widgets I recommend moving it to the context area. Might want to also review this @23Kev 🙏

Grollo13: Yes, I just realized when running the linter.

There is a lot of stuff going on in the fuel calculation, I'm trying to move the reusable code to the shared function bit by bit without braking it.

tariknz: To be safe do you want to just duplicate what you need and we can revisit it in a seperate PR?

Grollo13: Yes I think this might be the safer approach. So I'll add the lap counter using the new calculator, we can see how it performs, and then later we can update the fuel calculator.

Grollo13: @23Kev I've taken over part of your fuel calculator code in `useTotalRaceLaps.tsx`. Can you please review the part related to the leader car calculations? You are using `useEffect`, which as per my understanding executes the code inside only if the dependencies changed. I can't use it here though, as the return value of `useTotalRaceLaps()` is used inside a render function (`SessionBar.tsx, line 93`) which leads to linter errors (Refs cant be accessed during render).

src/frontend/context/shared/useTotalRaceLaps.tsx
```diff
@@ -0,0 +1,120 @@
+import { useSessionLapCount } from '../../components/Standings/hooks/useSessionLapCount';
+import {
+  useCurrentSessionType,
+  useFocusCarIdx,
+  useTelemetry,
+  useTelemetryValue,
+  useCarIdxClassEstLapTime
+} from '@irdashies/context';
+import { GlobalFlags } from '@irdashies/types';
```
tariknz:
```suggestion
```

Grollo13: fixed

PR #217, opened 2026-01-04
feat: Rejoin Indicator

Xenon1:
### Overview
[Dev ticket 4](https://discord.com/channels/1339669873386586152/1457043704803229882)

Added a rejoin indicator. This widget only shows when the player car drops below a user configured speed. It is also auto hidden when in the pits. 

User can define: 
- what speed (kmH) the widget will display at
- Gap where the care indicator is shown
- Gap where the stop indicator is shown

image

If the gap to the next car back is more than the Care gap, the clear indicator is shown
image

If the gap is equal too or less than the care gap, but more than the stop gap then the caution indicator is shown
image

If the gap is equal too or below the stop gap, the do not rejoin indicator is shown
image

### Testing
Tested in a AI practice session. 
- Altered all gaps and show at speed and changes were picked up. 
- Ensured widget was hidden when in the garage, in the pit stall and on pit road regardless of speed
- Ensured the widget only displayed cars on track and ignored cars in the pit lane

Note: When the user car comes to a stop, the widget may disapear for a second or 2 and come back. This is due to iRacing telemetry setting the car speed to a random value then back to 0

### DIFF
--- a/src/frontend/components/RejoinIndicator/RejoinIndicator.tsx
+++ b/src/frontend/components/RejoinIndicator/RejoinIndicator.tsx
@@ -4,7 +4,7 @@
  * Hidden until player is at a user defined speed (Default 30) and not in the pit lane. The gap thresholds are user configured as well
  */
 
-import { useTelemetryValues, useTelemetryValue, useFocusCarIdx, useCarIdxSpeed } from '@irdashies/context';
+import { useTelemetryValues, useTelemetryValue, useFocusCarIdx } from '@irdashies/context';
 import { useDriverRelatives } from '../Standings/hooks/useDriverRelatives';
 import { useRejoinSettings } from './hooks/useRejoinSettings';
 import type { Standings } from '../Standings/createStandings';
@@ -12,11 +12,10 @@ import type { Standings } from '../Standings/createStandings';
 export const RejoinIndicator = () => {
   const settings = useRejoinSettings();
   const playerIndex = useFocusCarIdx();
-  const telemetrySpeed = useTelemetryValues('Speed');
-  const carSpeeds = useCarIdxSpeed();
   const isInGarage = useTelemetryValue<number>('IsInGarage') === 1;
   const playerInPitStall = useTelemetryValue<number>('PlayerCarInPitStall') === 1;
   const carIdxOnPitRoad = useTelemetryValues<boolean[]>('CarIdxOnPitRoad');
+  const carSpeedForPlayer = useTelemetryValue('Speed');
 
   const drivers = useDriverRelatives({ buffer: 3 });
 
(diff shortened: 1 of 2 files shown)


src/frontend/components/RejoinIndicator/RejoinIndicator.tsx
```diff
@@ -0,0 +1,124 @@
...
+      } else {
+        carBehind = undefined;
+      }
+    }
+  }
+
+  // Read telemetry and computed car speed for the focused car index
+  const telemetryPerCar = telemetrySpeed ?? [];
+  const telemetrySpeedForPlayer = telemetryPerCar?.[playerIndex]; // likely m/s in live telemetry
+  const carSpeedForPlayer = carSpeeds?.[playerIndex]; // km/h when available
```
tariknz: Hey I think there's some confusion on how this works.

the telemetry value `Speed` has a single value and it is always the player's speed in m/s
`useCarIdxSpeed` does some internal computing to try to calculate all the car speeds based on their position on the track over time against the track length. There's a processing overhead to get this data so if you have no use for the speeds of other drivers you shouldn't bother using it and just stick with Speed.

'Speed' should be retrieved via `useTelemetryValue('Speed')` without the "s" to always get the first value in the array.

So I think this can be simplified to just 
```
const carSpeedForPlayer = useTelemetryValue('Speed');
const speedKmH = carSpeedForPlayer * 3.6;
```

src/frontend/components/Settings/sections/RejoinIndicatorSettings.tsx
```diff
@@ -0,0 +1,116 @@
...
+      enabled: savedSettings.enabled,
+      config: migrateConfig(savedSettings.config),
+    };
+
+    const prev = settingsRef.current;
+    if (prev.enabled === next.enabled && JSON.stringify(prev.config) === JSON.stringify(next.config)) return;
+
+    const id = setTimeout(() => setSettings(next), 0);
+    return () => clearTimeout(id);
+  }, [savedSettings]);
```
tariknz: This seems hacky? The other settings pages don't have this so why is it needed here? I think you can get rid of it.

```suggestion
```

src/frontend/components/RejoinIndicator/RejoinIndicator.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const speedKmH = Number.isFinite(carSpeedForPlayer)
+    ? carSpeedForPlayer
+    : Number.isFinite(telemetrySpeedForPlayer)
+    ? telemetrySpeedForPlayer * 3.6
+    : 0;
+
+  const gap = Math.abs(carBehind?.delta ?? Number.POSITIVE_INFINITY);
+  const gapLabel = Number.isFinite(gap) ? gap.toFixed(1) : '--';
+
+  // Exported helper below is used for unit testing the mapping
```
tariknz: No tests though?
Probably unecessary to have getStatusFromGap as a seperate function just inline it

PR #232, opened 2026-01-11
feat: pitlane helper

23Kev:
Features

- Speed Delta Display: Shows how far over/under the pit speed limit you are
  - Green: Safe (more than 5 km/h under limit)
  - Amber: Caution (0-5 km/h under limit)
  - Red: Over limit
  - Flashing Red Box: Severely over (more than 1.5 km/h over limit)
- Pitbox Distance: Countdown showing meters to your pitbox with progress bar
- Pit entry and exit locations detected and stored for each track
- Early Pitbox Warning: Alerts when your pitbox is near pit entry (last 10% of track)
- Pit Limiter Warning: Flashing warning when entering pit without limiter active
- Team Race Warning: Extra urgent warning after pit stop completion in team races
- Pitlane Traffic: Shows count of cars ahead and behind in pitlane

Modified/Added files

- src/app/storage/defaultDashboard.ts
- src/frontend/WidgetIndex.tsx
- src/frontend/components/PitlaneHelper/PitlaneHelper.stories.tsx
- src/frontend/components/PitlaneHelper/PitlaneHelper.tsx
- src/frontend/components/PitlaneHelper/hooks/usePitLimiterWarning.tsx
- src/frontend/components/PitlaneHelper/hooks/usePitSpeed.tsx
- src/frontend/components/PitlaneHelper/hooks/usePitboxPosition.tsx
- src/frontend/components/PitlaneHelper/hooks/usePitlaneHelperSettings.tsx
- src/frontend/components/PitlaneHelper/hooks/usePitlaneTraffic.tsx
- src/frontend/components/PitlaneHelper/hooks/usePitlaneVisibility.tsx
- src/frontend/components/Settings/SettingsLayout.tsx
- src/frontend/components/Settings/sections/PitlaneHelperSettings.tsx
- src/frontend/components/Settings/sections/index.ts
- src/frontend/components/Settings/types.ts

Ready for review and testing.


23Kev:
Still in development and testing so closing for now. Will reopen once its properly done.

src/frontend/context/PitLaneStore/usePitLaneDetection.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { usePitLaneStore } from './PitLaneStore';
+import { useTelemetryValues } from '../TelemetryStore/TelemetryStore';
+import { useSessionStore } from '../SessionStore/SessionStore';
+
+/**
+ * Hook that monitors telemetry and detects pit entry/exit positions.
+ * This should be mounted once at the app level to continuously track all cars.
+ */
+export const usePitLaneDetection = () => {
+  const session = useSessionStore((state) => state.session);
```
tariknz: Don't do this, if you request the session directly then the updates will happen at the rate of the session updates. If you get the trackID instead its memoised so it will only update when the sessionc hanges.

```suggestion
  const trackId = useSessionStore((state) => state.session?.WeekendInfo?.TrackID?.toString() ?? null);
```

src/frontend/context/PitLaneStore/usePitLaneDetection.tsx
```diff
@@ -0,0 +1,89 @@
...
+  useEffect(() => {
+    if (!trackId) {
+      reset();
+      return;
+    }
+
+    // Track changed
+    if (trackId !== currentTrackId) {
+      // Load from disk via IPC
+      window.electron?.pitLane?.getPitLaneData(trackId).then((data) => {
```
tariknz: What's the reason for saving and loading the pit entry/exit data into disk? I'm not understanding this here. Is it for performance? 

Has it actually been an issue?

src/frontend/App.tsx
```diff
@@ -46,6 +47,7 @@ const App = () => {
       <RunningStateProvider bridge={window.irsdkBridge}>
         <SessionProvider bridge={window.irsdkBridge} />
         <TelemetryProvider bridge={window.irsdkBridge} />
+        <PitLaneDetector />
```
tariknz: You want to pass the bridge like the above providers to not create a dependency on electron in the frontend. Best to follow the pattern for the irsdkBridge or dashboardBridge

src/main.ts
```diff
@@ -36,13 +37,16 @@ app.on('ready', async () => {
   const dashboard = getOrCreateDefaultDashboard();
   const bridge = getCurrentBridge();
 
+  // Setup IPC bridges
+  setupPitLaneBridge();
+
```
tariknz, line 43: I'm not super into having specific widget logic be in the main - there might be a better solution. Minor we can tidy this up later.

src/app/rendererExpose.ts
```diff
@@ -8,4 +9,14 @@ contextBridge.exposeInMainWorld('globalKey', {
...
+
+const pitLaneBridge: PitLaneBridge = {
+  getPitLaneData: (trackId: string) => ipcRenderer.invoke('pitLane:getData', trackId),
+  updatePitLaneData: (trackId: string, data: Partial<PitLaneTrackData>) =>
+    ipcRenderer.invoke('pitLane:updateData', trackId, data),
+};
+
+contextBridge.exposeInMainWorld('electron', {
+  pitLane: pitLaneBridge,
+});
```
tariknz: Any reason why this is called electron?

src/frontend/context/PitLaneStore/usePitLaneDetection.tsx
23Kev: This was done so each user had their own source of pit entry and exit locations that persisted across restarts, sessions etc. We could have this in a DB that the app called via api but that's a lot more maintenance and dependencies. 

If there is another way I'm all ears.

src/app/rendererExpose.ts
23Kev: My mistake, this has been changed.

src/frontend/App.tsx
23Kev: pitLaneProvider has been created as suggested

src/frontend/context/PitLaneStore/usePitLaneDetection.tsx
23Kev: Fixed - changed to use trackId selector: `useSessionStore((state) => state.session?.WeekendInfo?.TrackID?.toString() ?? null)` to prevent unnecessary re-renders.

src/frontend/App.tsx
23Kev: Fixed - created PitLaneProvider component that accepts bridge prop, following the same pattern as SessionProvider and TelemetryProvider. Updated App.tsx to use `<PitLaneProvider bridge={window.pitLaneBridge} />`

src/main.ts
23Kev: Acknowledged - left as-is for now since this can be tidied up later as noted.

23Kev:
@tariknz 

- Fixed usePitLaneDetection.tsx to use memoized trackId selector
- Created new PitLaneProvider.tsx component accepting bridge prop
- Updated App.tsx to use <PitLaneProvider bridge={window.pitLaneBridge} />
- Renamed window.electron to window.pitLaneBridge in rendererExpose.ts
- Updated window type definitions in interface.d.ts
- Removed obsolete PitLaneDetector.tsx component
- All changes verified with linting - passes with no errors

src/frontend/context/PitLaneStore/usePitLaneDetection.tsx
tariknz: Ok I see so we don't actually get this data from telemetry? 
Where would the users get the data from?

tariknz: Oh wait does it get set when you enter and exit the pit, the persist from there on?

23Kev: > Oh wait does it get set when you enter and exit the pit, the persist from there on?

Yep, this is what happens.
There is no pitlane entry/exit data in the telemetry. So the first time the driver exits and enters the Track% is stored in the pitLaneData.json file like below. If they never enter then we just store a null value and the smarts for pit entry (early pitbox detection) just use a default.
`{
  "9": {
    "pitEntryPct": 0.878361701965332,
    "pitExitPct": 0.9491437077522278
  },
  "47": {
    "pitEntryPct": 0.9199294447898865,
    "pitExitPct": 0.013613536022603512
  },
  "181": {
    "pitEntryPct": 0.9869002103805542,
    "pitExitPct": 0.06826776266098022
  }
}`

tariknz: Yeah makes sense, smart. Thanks for clarifying!

tariknz (approved):
LGTM

PR #234, opened 2026-01-11
Fix the regex to sanitize YAML input

Grollo13:
The regex was supposed to fix trailing commas, which it didn't actually
do.
Additionaly, iRacing telemetry can also publish lists with leading commas,
which will cause an error in the YAML parser.

The regex was updated to remove leading AND trailing regexes.

Validated with test-data on https://regex101.com/r/d6I24n/1


23Kev (commented):
This looks good and a lot more robust. I'll have to test it in an AI session later.

PR #235, opened 2026-01-11
Show lap counter in fixed lap races without decimals

Grollo13:
When in a fixed-lap race, the lap counter would show e.g. `20.0` total laps, which does not make sense if the number of laps is fixed and not estimated.

Also a minor bug was removed that would always take the drivers car lap, even if focussing on another car.


23Kev (approved):
This looks good to me

PR #240, opened 2026-01-13
[Bugfix] SOF calculation

gui-seppe:
Fix to the sof calculation. Test updated to reflect correct value


23Kev (approved):
These calculations look correct and result in the correct SOF values in sessions. Good to go.

PR #241, opened 2026-01-13
Feat/option for autostart after system start

creagel:
Implementation of settings for automatic startup after system boot


23Kev (approved):
Looks good, approving to merge.

Issue #28, opened 2025-06-14
[Standings] Deltas should indicate if the car is more than a lap down (e.g. display +2 laps)


tariknz:
You should be able to calculate laps to leader and figure out if they are on the lead lap else show how many laps they are down instead of the delta.

tariknz:
Closed by #244

PR #244, opened 2026-01-14
feat: Standing's gap shows laps count behind

BatMichoo:
Issue #28 Deltas should indicate if the car is more than a lap down

Standings now show lap count behind as -nL, as per the iRacing Results Tab.

interface Gap {
 value: number,
 laps: number 
} 

- Added above interface to not override the gap value with the lap count as interval uses said gap.
- Refactored createStandings.ts to use above interface
- Refactored augmentStandingsWithGap to calculate actual lap difference based on the gap value and class leader's fastest lap.
- Refactored DriverInfoRow to accept interface Gap.
- Refactored DeltaCell to accept interface Gap. If laps property is not 0 will show lap difference, otherwise actual gap.


tariknz (approved):
Its difficult to tell whats changed with the formatting changes - please just double check those for next time.

PR #245, opened 2026-01-14
Widgets show only on track

BatMichoo:
Fuel Calculator
Blind Spotter
Rejoin
Faster Cars

are missing the option to be hidden while not in the car.

All 4 have been updated to do so.

Updated each of the overlays settings and added a ToggleSwitch in the UI. Updated the settings hooks as well.


23Kev:
@BatMichoo I did see you made some changes to the FuelCalculator. Are they regarding the same thing or did you mean to have them in a different PR?

BatMichoo:
Apart from the showOnlyWhenOnTrack,  no I think prettier is just formatting the whole thing differently

PR #247, opened 2026-01-15
fix: Lapped car state (blue/red highlighting) is incorrect when crossing the line

Grollo13:
To determine wheter a car has lapped another car, it is not sufficient
to just compare the lap numbers. If we do, starting a new lap will let
cars behind appear as lapped until they also started a new lap.
To accuratly determine the lapped state, we have to add the lap number and
lap distance percentage, substract one from the other, and round it to
the next integer value.

Example1:
Car1 started lap 5, Car2 just behind did not yet cross the line.
Without correction:
    `5 - 4 = 1` => Car2 is one lap behind Car1
With correction:
    `Math.round((5 + 0.1) - (4 + 0.9)) = 0.2` => Car1 did not lap Car2

Example2:
Car1 approaches Car2, which is a lap behind.
    `Math.round((5 + 0.4) - (4 + 0.5)) = 0.9` => Car1 is about to lap Car2

Car1 lapped Car2, and is ahead on the track
    `Math.round((5 + 0.7) - (4 + 0.6)) = 1.1` => Car1 has lapped Car2

### DIFF
--- a/src/frontend/components/Standings/hooks/useDriverPositions.tsx
+++ b/src/frontend/components/Standings/hooks/useDriverPositions.tsx
@@ -10,6 +10,7 @@ import {
   usePrevCarTrackSurface,
   useFocusCarIdx,
   useSessionPositions,
+  useTelemetryValues,
 } from '@irdashies/context';
 
 import { Standings, type LastTimeState } from '../createStandings';
@@ -39,6 +40,7 @@ export const useDriverPositions = () => {
   const prevCarTrackSurface = usePrevCarTrackSurface()
   const lastPitLap = usePitLap()
   const lastLap = useCarLap()
+  const carIdxLapDstPct = useTelemetryValues('CarIdxLapDistPct');
 
 
   const positions = useMemo(() => {
@@ -51,6 +53,7 @@ export const useDriverPositions = () => {
       lastLap: lastLap[carIdx] ?? -1,
       lastLapTime: carIdxLastLapTime?.value?.[carIdx] ?? -1,
       lapNum: carIdxLapNum?.value?.[carIdx],
+      lapDstPct: carIdxLapDstPct[carIdx] ?? 0,
       lastPitLap: lastPitLap[carIdx] ?? undefined,
       prevCarTrackSurface: prevCarTrackSurface[carIdx] ?? undefined,
       carTrackSurface: carIdxTrackSurface?.value?.[carIdx]
@@ -63,6 +66,7 @@ export const useDriverPositions = () => {
     lastLap,
     carIdxF2Time?.value,
     carIdxLapNum?.value,
+    carIdxLapDstPct,
     lastPitLap,
     prevCarTrackSurface,
     carIdxTrackSurface?.value
@@ -152,6 +156,9 @@ export const useDriverStandings = () => {
     const playerLap = playerCarIdx !== undefined
       ? driverPositionsByCarIdx.get(playerCarIdx)?.lapNum ?? 0
       : 0;
+    const playerLapDistPct = playerCarIdx !== undefined
+      ? driverPositionsByCarIdx.get(playerCarIdx)?.lapDstPct ?? 0
+      : 0;
 
     const standings = drivers.map((driver) => {
       const driverPos = driverPositionsByCarIdx.get(driver.carIdx);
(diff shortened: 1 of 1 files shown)


src/frontend/components/Standings/hooks/useDriverPositions.tsx
```diff
@@ -34,6 +34,7 @@ export const useDriverPositions = () => {
   const carIdxLastLapTime = useTelemetry('CarIdxLastLapTime');
   const carIdxF2Time = useTelemetry('CarIdxF2Time');
   const carIdxLapNum = useTelemetry('CarIdxLap');
+  const carIdxLapDstPct = useTelemetry('CarIdxLapDistPct')
```
tariknz:
```suggestion
  const carIdxLapDstPct =  useTelemetryValuesMapped<number[]>(
    'CarIdxLapDistPct',
    (val) => Math.round(val * 100) / 100 // Reduce precision to 1 decimal places to minimize unnecessary updates
  );
```
I would try to map this to a decimal point - the values you get from this is something like 0.9942072629928589

If you keep it like this that means this whole hook will recompute at the full update rate (25hz currently)

tariknz: I'd honestly even make it less percise like all you care about is the crossing from 0.999 to 0

Grollo13: Right, I've reduced it to one decimal

I'll test it later today

tariknz: Thanks! let me know once you get it tested. Other than that code-wise is good for me

Grollo13: There are some glitches when the car distance is ~0.5 laps. Lapped cars might already appear ahead in the relative, but are still not highlighted as blue for a short time.

Grollo13: I think it mainly comes down to the fact that `useDriverRelatives` and `useDriverPositions` use different precisions.
Also, I observed a case in which the lap number is incremented before lapDistPct is reset to 0

Grollo13: @tariknz 

I tried now by reducing the precision for both to 0.0001 => 0.01%
On the Nürburgring 24h layout (~25km), this equals to 2,5m.
At the full update rate of 25Hz, a car has to travel 62,5 m/s or 225km/h to go at least 2,5m per tick.
So theoretically, we could save a bit of runtime for each car that is slower.
But we also have to consider that we use the values of ALL cars, so at least one car moving across the 2,5m rounding boundary will trigger a new calculation. With dozens of cars, this will realistically happen every tick anyway.

Rounding to 2,5m for relative positions is also not good enough, as there are a lot of glitches (drivers moving up and down the order of the relative constantly, without actually changing positions) especially in the first lap, but also for any kind of close position fights. 
I think for the relative, we must use the highest precision to get a clear and stable relative.

Now for the lapped state, as mentioned above, once we reduce the precision, we get glitches if it is not synchronized with the relative. Users will see a new car showing up on the relative first as white, then it will turn blue, maybe white again, going back and forth until it is near enough that the rounding does not make a difference anymore.

I tried to analyze the CPU profiles, but I'm struggling a bit with this. Claude also was not very helpful. But I did a very basic check by observing the FPS at a standing race start. Once the cars start moving, the relative will get updated constantly, as cars move up the lap distance. Here at least I could see no FPS drops. 
A proper CPU profile analysis would be nicer of course, but at least we dont see any performance impact like we saw for the pit info or weather widget in the current release.

tariknz: Yep ok fair enough
You happy to have this merged in its current state?

tariknz: In regards to the CPU profile I think its not really helping me either I am not getting a great output from it. I am thinking we might need to run the dev tools profiling instead.

tariknz (approved):

src/frontend/components/Standings/hooks/useDriverPositions.tsx
Grollo13: Right, lets merge then!

PR #249, opened 2026-01-15
feat: compact mode

LanceVShoot:
Compact mode general settings option to remove spacing from standings/relatives for a more visually compact view.

merged-image


tariknz (approved):
Just when you get the chance would be good to add storybook stories for compact mode for standings/relative.

PR #251, opened 2026-01-16
fix: widgets show only on track rejoin indicator and faster cars behind

BatMichoo:
Removed logic from Rejoin Indicator as it was already present.

Added missing check for the Faster Cars overlay.


tariknz (approved):
LGTM

PR #255, opened 2026-01-16
fix: drivers behind customizations

LanceVShoot:
For showing drivers behind, don't show drivers that are in the pits. Also allow showing multiple drivers behind, and adjust the alignment/order of the drivers within the widget.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "irdashies",
-  "version": "0.0.37",
+  "version": "0.0.38",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "irdashies",
-      "version": "0.0.37",
+      "version": "0.0.38",
       "license": "MIT",
       "dependencies": {
         "@phosphor-icons/react": "^2.1.10",


23Kev:
## Review: Drivers Behind Customizations

Great enhancement to the Faster Cars From Behind widget! The pit lane filtering and multiple driver support are excellent additions. However, there's one critical issue that should be fixed before merging.

---

### Critical Issue: React Key Using Array Index

**Current code (Line 32 in FasterCarsFromBehind.tsx):**
```typescript
{orderedCars.map((car, index) => (
  <FasterCarsFromBehindDisplay key={index} {...car} />
))}
```

**Problem:** Using array index as the `key` is problematic when:
- The array order changes (which happens with the `Reverse` setting)
- Items are added/removed (when cars pit/unpit)

This can cause React to incorrectly track components, leading to:
- Incorrect animations
- Stale data in components  
- Performance issues

**Recommended Fix:**

1. Add `carIdx` to the data returned from `useCarBehind`:

```typescript
// In useCarBehind.tsx
return carsBehind
  .filter(/* ... */)
  .map(car => ({
    carIdx: car.carIdx,  // ← Add this
    name: car.driver?.name,
    distance: parseFloat(car.delta?.toFixed(1) ?? '0'),
    classColor: car.carClass?.color,
    percent: percent,
  }));
```

2. Update the interface:

```typescript
// In FasterCarsFromBehind.tsx
export interface FasterCarsFromBehindProps {
  carIdx?: number;  // ← Add this
  name?: string;
  distance?: number;
  percent?: number;
  classColor?: number;
}
```

3. Use `carIdx` as the key:

```typescript
{orderedCars.map((car) => (
  <FasterCarsFromBehindDisplay key={car.carIdx} {...car} />
))}
```

---

### Suggestions for Improvement

#### 1. Setting Labels Could Be Clearer

**Current:** "Closest Driver" with options "Top" / "Reverse"

**Suggestion:** Make it more descriptive:
```typescript
<span className="text-sm text-slate-300">Closest Driver Position</span>
<option value="Top">Closest at Top</option>
<option value="Reverse">Closest at Bottom</option>
```

This makes it clearer what each option does.

#### 2. Potential Confusion: Two Different "Top" Settings

Currently both settings use "Top" but mean different things:
- **Align Driver Boxes: Top** = align boxes to top of widget container
- **Closest Driver: Top** = closest driver appears in top position

This might confuse users. Consider:
- More distinct labels
- Combining into a single setting
- Adding tooltip/help text

#### 3. Increase Gap Between Driver Boxes

Current gap is `gap-1` (4px). Consider using `gap-2` (8px) for better visual separation when showing multiple drivers:

```typescript
<div className={`flex flex-col gap-2 h-full ${containerAlignment}`}>
```

---

Overall excellent work! Just fix the React key issue and this will be ready to merge.

tariknz (approved):

PR #257, opened 2026-01-17
fix: Relative Widget Position Display Bugs

Xenon1:
## Issues Fixed:

- Single-class off-by-one error: Relative widget was displaying positions 1 higher than actual (e.g., showing position 23 instead of 22)
### Before
image

### After
image

- Multiclass position bug: Relative widget was showing overall position instead of in-class position in multiclass races
### Before
image

### After
image

## Testing
- tested in AI practice session and PCC practice session
- test and lint passed


tariknz:
This PR doesn't really match the description. It sounded like the positions were off-by-one but the PR forces live positions when in a multiclass race? 

I am not super comfortable with the state of live positions currently so I don't want to default it to be enabled at this stage.

PR #258, opened 2026-01-17
fix: focus car for session laps

LanceVShoot:
session laps hook was using player car laps so it was showing 0 when spectating. using focus car to see actual lap count. (in team event, and/or replay during race while pitting, etc)


tariknz (approved):
LGTM

PR #260, opened 2026-01-18
fix: move outlap badge to the pitcell

tariknz:
image

The outlap badge is related to the pit status it doesn't make sense for it to be in the driver name. It also causes both badges to appear but far from each other.


djdiliberti:
Hey, I did this change initially. I agree that outlap is related to pit status, but I found it annoying that it resized the entire column to accommodate the outlap badge. I don't keep pit status in the default location so it creates a lot of movement in the table. Maybe we can do something like if the person is on an outlap, don't display the pit lap during the outlap? This way it wouldn't force a big resize anytime someone pits. On the outlap the lap pitted is kind of extraneous information since we know what lap we're on, so maybe no one will miss it.

tariknz:
@djdiliberti what do you think of this? On outlap we show the time if available but not the pit lap number. So only one badge appears during an outlap

image

djdiliberti:
I think that's pretty good.

PR #271, opened 2026-01-20
feat: Add option to select display mode for pit lap

Grollo13:
The pit lap cell in the relative/standings currently shows the lap of the last pit stop for a car. This information might not be useful in longer and/or multiclass races, as the current lap can be very different for the cars.

This PR adds an option to the relative/standings configuration to select to either show the lap as mentioned above, or the number of laps since the last pit stop. This will make it easier to check whether a car is at the beginning, middle, or end of a stint

grafik

grafik

### DIFF
--- a/src/app/storage/defaultDashboard.ts
+++ b/src/app/storage/defaultDashboard.ts
@@ -53,7 +53,8 @@ export const defaultDashboard: DashboardLayout = {
           "enabled": false
         },
         "pitStatus": {
-          "enabled": true
+          "enabled": true,
+          "pitLapDisplayMode": "lapsSinceLastPit"
         },
         "position": {
           "enabled": true
@@ -239,7 +240,8 @@ export const defaultDashboard: DashboardLayout = {
           "enabled": false
         },
         "pitStatus": {
-          "enabled": true
+          "enabled": true,
+          "pitLapDisplayMode": "lapsSinceLastPit"
         },
         "carManufacturer": {
           "enabled": true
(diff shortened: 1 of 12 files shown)


src/app/storage/defaultDashboard.ts
```diff
@@ -53,7 +53,8 @@ export const defaultDashboard: DashboardLayout = {
           "enabled": false
         },
         "pitStatus": {
-          "enabled": true
+          "enabled": true,
+          "pitLapDisplayMode": "lapsSinceLastPit"
```
tariknz, line 57: Can we make the default what it was, so lastPitLap. I feel like that should be the default.
But maybe I'm wrong feel free to raise the question in Discord.

tsconfig.json
```diff
@@ -17,7 +17,8 @@
       "@irdashies/storybook": ["./.storybook"],
       "@irdashies/utils/*": ["./src/frontend/utils/*"],
       "@irdashies/context": ["./src/frontend/context"],
-      "@irdashies/types": ["./src/types"]
+      "@irdashies/types": ["./src/types"],
+      "@irdashies/settings": ["./src/frontend/components/Settings"]
```
tariknz: What did you need this alias for? It doesn't seem to be used.

src/app/storage/defaultDashboard.ts
```diff
@@ -239,7 +240,8 @@ export const defaultDashboard: DashboardLayout = {
...
```
tariknz, line 244: Same here

src/app/storage/defaultDashboard.ts
Grollo13: That was also my first intention, but I really think that most users would actually prefer this setting. We could start a poll in the discord maybe

tsconfig.json
Grollo13: Ah yes, left-over. I'll remove it

Grollo13: dropped the commit

PR #279, opened 2026-01-21
chore: use time utility for pit stop duration

LanceVShoot:
pit stop duration to show as:

21s = :21
1m21s = 1:21


tariknz:
Do you find it weird seconds doesn't start with 0? e.g 0:23

PR #288, opened 2026-01-24
chore: performance improvements in the sdk

tariknz:
- Updates input trace to use refs to avoid unnecessary rerenders of the whole window
- Move a function for the pitlane helper to outside the component to avoid it being recreated
- Add telemetry caching to avoid copying the telemetry data every render
- Add session caching to only publish via IPC when the session data version changes
- Avoid yaml parsing when the data version hasn't changed

### DIFF
--- a/src/app/irsdk/native/irsdk_node.cc
+++ b/src/app/irsdk/native/irsdk_node.cc
@@ -113,7 +113,6 @@ Napi::Value iRacingSdkNode::WaitForData(const Napi::CallbackInfo &info)
     timeout = info[0].As<Napi::Number>();
   }
 
-  // Try to start SDK if not connected
   if (!irsdk_isConnected() && !irsdk_startup()) {
     return Napi::Boolean::New(info.Env(), false);
   }


.claude/settings.local.json
```diff
@@ -5,7 +5,13 @@
...
       "Bash(git checkout:*)",
       "Bash(npx tsc:*)",
-      "Bash(npm test:*)"
+      "Bash(npm test:*)",
+      "Bash(while read f)",
+      "Bash(do head -30 \"$f\")",
+      "Bash(echo:*)",
+      "Bash(done)",
+      "Bash(xargs cat:*)",
+      "Bash(ls:*)"
```
tariknz:
```suggestion
      "Bash(npm test:*)"
```

src/app/irsdk/native/irsdk_node.cc
```diff
@@ -113,53 +113,66 @@ Napi::Value iRacingSdkNode::WaitForData(const Napi::CallbackInfo &info)
     timeout = info[0].As<Napi::Number>();
   }
 
+  // Try to start SDK if not connected
```
tariknz:
```suggestion
```

PR #290, opened 2026-01-24
feat: Bold text in the overlays

gui-seppe:
I've seen that bold text is used in a lot of other overlays, so I added a toggle to turn on/off bold text in the overlays.
Imo it makes the text more readable, especially when using the medium font size in the relative and standings widgets, without using larger fonts, which take up more screen space.

The toggle is located in the 'General' tab
image
Some screenshots (medium font used)
image
image
image
image

I tested it in a practice session and didn't notice any issues, but a double check does not hurt.

### DIFF
--- a/src/frontend/components/ThemeManager/ThemeManager.stories.tsx
+++ b/src/frontend/components/ThemeManager/ThemeManager.stories.tsx
@@ -31,12 +31,12 @@ const createMockBridge = (
   saveDashboard: (dashboard: DashboardLayout) => {
     setFontSize(dashboard.generalSettings?.fontSize);
     setColorPalette(dashboard.generalSettings?.colorPalette || 'default');
-    setFontWeight(dashboard.generalSettings?.fontWeight);
+    setFontWeight(dashboard.generalSettings?.fontWeight || 'normal');
   },
   dashboardUpdated: (callback) => {
     callback({
       widgets,
-      generalSettings: { fontSize, fontWeight, colorPalette },
+      generalSettings: { fontSize, colorPalette, fontWeight },
     });
     return () => {
       return;
@@ -53,7 +53,7 @@ const createMockBridge = (
   resetDashboard: () =>
     Promise.resolve({
       widgets: [],
-      generalSettings: { fontSize, colorPalette, fontWeight },
+      generalSettings: { fontSize, colorPalette },
     }),
   toggleDemoMode: () => {
     return;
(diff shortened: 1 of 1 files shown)


src/frontend/theme.css
```diff
@@ -2,6 +2,11 @@
 layer(base);
 @import 'tailwindcss';
 
+/* Toggle bold for all overlay text when class is applied */
+.overlay-window.bold-text * {
+  font-weight: 800;
+}
```
tariknz: This should be:

```
.overlay-window.overlay-theme-font-weight-bold {
  font-weight: var(--font-weight-bold);
}
```

src/frontend/App.tsx
```diff
@@ -19,20 +19,31 @@ import { HideUIWrapper } from './components/HideUIWrapper/HideUIWrapper';
...
+                  className={`overlay-window ${
+                    generalSettings?.boldText ? 'bold-text' : ''
+                  }`}
+                >
+                  <WidgetComponent {...widget.config} />
+                </div>
+              ) : (
+                <></>
+              )
+            }
```
tariknz: Use the ThemeManager.tsx to apply theme classes

See existing patterns please for the text size and colour

src/types/dashboardLayout.ts
```diff
@@ -35,6 +35,7 @@ export interface GeneralSettingsType {
   enableAutoStart?: boolean;
   compactMode?: boolean;
   overlayAlwaysOnTop?: boolean;
+  boldText?: boolean;
```
tariknz: I prefer not to have a boolean, then later we're going to have someone add another weight what will we do?

For now make the type: `normal`, `bold` or `extrabold` if you want to use the 900 variant

These are the theme variables currently available, don't implement all the size but just use the same naming pattern:
```
  --font-weight-thin: 100;
  --font-weight-extralight: 200;
  --font-weight-light: 300;
  --font-weight-normal: 400;
  --font-weight-medium: 500;
  --font-weight-semibold: 600;
  --font-weight-bold: 700;
  --font-weight-extrabold: 800;
  --font-weight-black: 900;
```

https://tailwindcss.com/docs/theme

src/frontend/App.tsx
tariknz: Make sure to update the story also (see ThemeManager.stories.tsx) this will let you quickly visualise all the styling configs.

Lastly just ensure to have a quick test on the browser components make sure those render correctly.
e.g. http://localhost:3000/component/standings

PR #297, opened 2026-01-25
refactor: separate tachometer into independent widget

sdcarter:
- Extract tachometer/RPM LED functionality from Input widget into standalone Tachometer widget
- Create new Tachometer component with independent positioning and scaling
- Add TachometerSettings panel with custom shift points configuration
- Update Input widget to remove tachometer rendering
- Add tachometer widget to default dashboard (disabled by default)
- Update mock data to use Mazda MX-5 Cup for demo mode
- Add demo mode indicators showing which car is used in demo
- Remove tachometer configuration from Input settings

Potential breaking change: Users will need to enable and configure the new Tachometer widget separately from Input widget


src/app/irsdk/node/utils/mock-data/session.json
```diff
@@ -3413,15 +3413,15 @@
...
         "CarID": 122,
         "CarIsPaceCar": 0,
         "CarIsAI": 0,
         "CarIsElectric": 0,
-        "CarScreenName": "BMW M4 GT4",
-        "CarScreenNameShort": "BMW M4 GT4",
-        "CarClassShortName": "BMW M4 GT4",
+        "CarScreenName": "Mazda MX-5 Cup",
+        "CarScreenNameShort": "Mazda MX-5 Cup",
+        "CarClassShortName": "MX5",
```
tariknz: Did you mean to update this? I assume you were just testing it?

sdcarter: ahh. yeah. I was testing and never went back. Let me revert back to the M4 GT4 (as I think that is what it was before). Makes sense to keep it GT4 as that is now, basically, a legacy car so it's not bad to test with it.

PR #301, opened 2026-01-26
feat: add throttling for telemetry updates in live position calculation

RenatoCoicev:


src/frontend/context/shared/useThrottledValue.tsx
tariknz, line 1: All shared hooks live under context so just put it under frontend/context/shared export it in index and import it via @irdashies/context

PR #337, opened 2026-02-10
feat: adjustable font size for track maps

spikethedog78:
## Description

Adds config option to adjust the relative font size on the track map and flat track map

## Screenshots

image

### Before
image

### After
image

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

- [X] I have tested this in iRacing (either in an online session or with AI)
- [X] I have added/updated Storybook stories for visual changes
- [X] I have updated defaultDashboard.ts if introducing new widgets or configurations (if applicable)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "irdashies",
   "productName": "irdashies",
-  "version": "0.0.39",
+  "version": "0.0.40",
   "description": "iRacing Dashboards & Overlays",
   "repository": {
     "type": "git",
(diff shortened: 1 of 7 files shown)


src/frontend/components/Settings/sections/TrackMapSettings.tsx
```diff
@@ -156,6 +158,26 @@ export const TrackMapSettings = () => {
             </p>
           </div>
 
+          <div className="space-y-2">
+            <label className="text-slate-300">
+              Relative Font Size: {settings.config.trackmapFontSize ?? 100}%
```
tariknz, line 163: Just wondering if instead of it having its own setting we would jsut base it off the default font-size settings in general. We should be able to get the value to set in canvas via getting the css vars.

spikethedog78: I used a separate value as the track font is relative the size of the circle. You might want a small circle but big font, or you might want a larger circle with a smaller font.

spikethedog78: I've merged the changes with the new performance updates.

PR #340, opened 2026-02-11
feat: add Twitch Chat overlay widget

23Kev:
## Description

Add a new Twitch Chat overlay widget that displays live Twitch chat messages on screen. Uses the `@tmi.js/chat` library for read-only channel connection (no OAuth required).

Also introduces an `alwaysEnabled` widget property that allows widgets to remain visible when iRacing is not running, since Twitch Chat does not require telemetry data.

Based on work by @creagel, rebased cleanly onto current main.

### Features
- Live Twitch chat messages displayed as an overlay
- Configurable channel name, font size, and background opacity
- Widget stays visible when iRacing is not running (`alwaysEnabled`)
- Settings panel with channel name input, font size slider, and opacity slider

### Before
No Twitch Chat widget available.

### After
image

## Type of Change

## Checklist


package.json
```diff
@@ -87,6 +87,7 @@
   "license": "MIT",
   "dependencies": {
     "@phosphor-icons/react": "^2.1.10",
+    "@tmi.js/chat": "^0.7.4",
```
tariknz, line 90: Just looking at this dep it seems to be relatively new and unknown 

Image

Is this the only one available?

creagel: It is the successor to the original library (after the Twitch API change).

PR #341, opened 2026-02-11
feat: additional font options and weights

spikethedog78:
## Description

More granular font options. Includes a choice of font, more granular size options, and more weight options.

## Screenshots

image

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

- [X] I have updated defaultDashboard.ts if introducing new widgets or configurations (if applicable)

### DIFF
--- a/src/frontend/components/Settings/sections/GeneralSettings.tsx
+++ b/src/frontend/components/Settings/sections/GeneralSettings.tsx
@@ -93,20 +93,6 @@ export const GeneralSettings = () => {
       currentDashboard?.generalSettings?.enableNetworkAccess ?? false,
   });
 
-  const restrictedWeights =
-    FONT_WEIGHT_RESTRICTIONS[settings.fontType ?? ''] ?? [];
-
-  const filteredWeights = Object.fromEntries(
-    Object.entries(FONT_WEIGHT_PRESETS).filter(
-      ([key]) => !restrictedWeights.includes(key)
-    )
-  );
-
-  const resolvedFontWeight =
-  settings.fontWeight && filteredWeights[settings.fontWeight]
-    ? settings.fontWeight
-    : 'normal';
-
   if (!currentDashboard || !onDashboardUpdated) {
     return <>Loading...</>;
   }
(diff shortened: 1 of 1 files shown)


tariknz:
Hey what's your discord handle?
We'll need storybook updated for any visual change

src/frontend/components/Settings/sections/ProfileSettings.tsx
```diff
@@ -91,63 +42,6 @@ export const ProfileSettings = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  // Sync pending values when profile changes
```
tariknz, line 94: Why remove these? Can you please add some details to all changes

spikethedog78: When adding the new font options I found these didn't seem to do anything? I can re-add if I am wrong.

tariknz: Hmm I genuinely don't know

tariknz: I'll need to take a closer look this weekend if this can wait please

spikethedog78: Of course. Happy to contribute.

tariknz: Yep you're right this did nothing, good to remove it

src/frontend/theme.css
```diff
@@ -1,11 +1,21 @@
-@import url('https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,100;0,300;0,400;0,700;0,900;1,100;1,300;1,400;1,700;1,900&display=swap')
-layer(base);
```
tariknz, line 2: Why remove layer(base)? this might be needed by tailwind. I think it allows it to be applied to the base layer.

tariknz: I think this actually caused a bug the font seemed to have changed by default

spikethedog78: I've restored layer(base) to the default Lato font

src/frontend/theme.css
```diff
@@ -1,11 +1,22 @@
 @import url('https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,100;0,300;0,400;0,700;0,900;1,100;1,300;1,400;1,700;1,900&display=swap')
```
tariknz, line 1: I think we might be missing some weights for Lato
normal/medium is the same
and semi-bold and bold is the same

spikethedog78: Lato doesn't support all 5 font weights on Google Fonts. We could replace it with something that does?

tariknz: Ah ok boohoo.. do you think maybe we should reduce the bold options then? Just have the three, normal, semibold, bold?

I like Lato I don't want to suddenly change the look and feel it will be drastic. Then I am not looking forward to people raising bugs about it since its the default font 🫠

spikethedog78: OK, let me take another look.

spikethedog78: I've removed the weight options that are unavailable per font

tariknz (approved):

PR #485, opened 2026-04-11
Invalid lap times on race start laps under safety car

BatMichoo:
## Description

I ran a league that had safety cars and of course first one was on lap 1 turn 2. Couple of laps followed that were in the 4 min range. Normal pace is sub 2. Up until the restart the delta was going crazy as pace was mostly inconsistent (throttle / gas) on the fastest lap under safety car. Now if we have a saved lap time if the newer time is more that 15 % slower it will get disregarded completely.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist


BatMichoo:
I guess since I've not pulled in 3 weeks I've got some phantom changes entangled. I touched last 2 files only. The other 13 are formatting only.

PR #487, opened 2026-04-13
feat: resize slider in standings widget

toba-k:
## Description

Added a size slider in the Standings widget

## Screenshots

size_slider

### After

After 150% zoom applied:
zoomed_view_150

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a Styling setting to adjust the Standings widget scale from 50% to 200%.
  * Scale adjustments can be set in 5% increments and default to 100%.
  * The selected scale is saved and applied to the widget.

### DIFF
--- a/src/frontend/components/Settings/sections/StandingsSettings.tsx
+++ b/src/frontend/components/Settings/sections/StandingsSettings.tsx
@@ -808,7 +808,7 @@ export const StandingsSettings = () => {
                         getItemConfig={(id) => {
                           const item =
                             settings.config.headerBar[
-                            id as keyof typeof settings.config.headerBar
+                              id as keyof typeof settings.config.headerBar
                             ];
                           if (
                             typeof item === 'object' &&
@@ -822,7 +822,7 @@ export const StandingsSettings = () => {
                         updateItemConfig={(id, config) => {
                           const item =
                             settings.config.headerBar[
-                            id as keyof typeof settings.config.headerBar
+                              id as keyof typeof settings.config.headerBar
                             ];
                           if (
                             typeof item === 'object' &&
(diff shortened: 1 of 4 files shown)


src/frontend/components/Settings/sections/StandingsSettings.tsx
```diff
@@ -946,6 +946,24 @@ export const StandingsSettings = () => {
...
+                      min={100}
+                      max={300}
+                      step={1}
+                      onChange={(v) =>
+                        handleConfigChange({
+                          stylingOptions: {
+                            ...settings.config.stylingOptions,
+                            size: v,
+                          },
+                        })
```
toba-k, line 963: Not sure this is a good practice, based on what I saw in other files

tariknz:
For this one wouldn't it be better to just update the font-size instead? Everything usually scales around it so you don't need to zoom. e.g. have an override seperate to the general setting font size

toba-k:
> For this one wouldn't it be better to just update the font-size instead? Everything usually scales around it so you don't need to zoom. e.g. have an override seperate to the general setting font size

There are a lot of requests on Discord about making each widget re-sizeable, I guess this solution will become obsolete fast if we go for a "proper" solution. Or do you think this still would be a good addition for the time being?

tariknz:
> > For this one wouldn't it be better to just update the font-size instead? Everything usually scales around it so you don't need to zoom. e.g. have an override seperate to the general setting font size
> 
> There are a lot of requests on Discord about making each widget re-sizeable, I guess this solution will become obsolete fast if we go for a "proper" solution. Or do you think this still would be a good addition for the time being?

Ok fair enough as long as you have a good pattern that's transferrable for other widgets - can we call it scale or zoom? just thinking something more obvious - size is a little ambigious

tariknz (approved):

PR #488, opened 2026-04-13
feat: add sector delta widget

smithdt:
## Description

Ticket-0058

New Widget - Sector Deltas

- Displays delta to your session-best sector time or ghost lap (configurable)
- Displays a color-coded bottom border accent per sector (purple/green/yellow/red)
- In-sector progress bar on the current sector card showing % through the sector
- Ghost icon indicator when comparing against a ghost lap
- `ghostComparison` setting: 'Ghost When Available' or 'Session Best Only'
- Live running delta for the **current sector** vs ghost file (when loaded) or personal session best — uses PCHIP interpolation from the reference lap for accuracy

Updated Widget - Track Map

- Shares data from Sector Deltas
- Colors track sectors based on set thresholds compared to current session or ghost lap
- Added small dividers between sectors

Related changes

- Modified test data to include slight variance per lap (+/- 1%)
- Added Ghost telemetry data for ghost comparisons in this change and other future use
- Moved `useReferenceLapSectorTimes` to shared context for reuse across widgets
- Refactored `useLiveSectorDelta` to use `calculateReferenceDelta` from the Standings interpolation module, removing duplicate `interpolateAtPoint` calls

## Screenshots

image

## Type of Change

- [ ] Performance improvement
- [x] Refactoring (no functional changes)
- [ ] Documentation update
- [ ] Dependency update

## Checklist


dbadashboard:
Ticket-0058

New Widget - Sector Deltas

- Displays delta to your session-best sector time
- Displays a color based on % off-pace (configurable in menu)
- Support for relative performance to Ghost sector performance

Updated Widget - Track Map

- Shares data from Sector Deltas
- Colors track sectors based on set thresholds compared to current session
- Added small dividers between sectors

dbadashboard:
## Bug fix: active reset corrupting sector delta display

Investigated a bug where sector deltas showed wildly incorrect values (~10s off) after an active reset on a 3-sector track.

### Root cause

The wrap-around handler (S/F crossing) **unconditionally** overwrote `previousLapSectorTimes` with whatever was in `currentLapSectorTimes` — even when `currentLapSectorTimes` was entirely null. This happened on the "dummy lap" the player must drive after a reset before timing resumes.

Specifically:
1. Player crashes in sector 2 after completing sectors 0 and 1 → `currentLapSectorTimes = [S0, S1, null]`
2. Active reset → `invalidateLap()` fires; partial times kept in `currentLapSectorTimes` (intentional — valid sector times from invalid laps should be retained)
3. Player drives dummy lap (timing paused) and crosses S/F → partial times `[S0, S1, null]` correctly promoted to `previousLapSectorTimes` ✓
4. Player does a **second reset** before completing any sectors → `currentLapSectorTimes = [null, null, null]`
5. Player drives another dummy lap and crosses S/F → `previousLapSectorTimes` overwritten with all-null, **wiping the good reference data** ✗

After step 5, the display showed `--` or used session-best comparison against stale values, producing large incorrect deltas.

### Fix

In the wrap-around handler, `previousLapSectorTimes` is now only updated when `completedLapTimes` contains at least one non-null value:

- **Has valid partial times** (sectors 0 and 1 completed before crash): promoted to `previousLapSectorTimes` as before — valid partial data is preserved as display fallback on the next real lap
- **All-null dummy lap** (second reset before completing any sectors): existing `previousLapSectorTimes` is kept unchanged

`sessionBestSectorTimes` was not affected by this bug — those are committed at sector crossing time while `lapStarted = true` and were already correctly preserved across resets.

Two new tests added covering both scenarios.

dbadashboard:
## Latest changes (2026-04-17)

### Sector timing redesign — `lapStarted` → `sectorEntryValid`

The `lapStarted` global flag has been replaced with a per-sector `sectorEntryValid` boolean. This fixes a class of bugs where active resets would clear **all** sector reference data instead of only the sector the player teleported into.

**How it works:**
- A teleport/reset (detected by position delta or speed > 0.08 lap%/s) only invalidates the sector landed in — `currentLapSectorTimes[newSectorIdx] = null`. All other recorded sector times are preserved.
- `previousLapSectorTimes[i]` is now updated **immediately when each sector is validly completed** (not just at S/F crossing), so reference data survives any sequence of resets.
- `invalidateLap()` (called on `IsOnTrack → false`) only sets `sectorEntryValid = false` — it no longer touches timing arrays.
- The S/F crossing always sets `sectorEntryValid = true` for sector 0, so timing resumes naturally without needing a full warm-up lap.

**Teleport detection:** Added speed-based check (`delta / timeDelta > 0.08 pct/s`) alongside the existing absolute threshold. This catches active resets on 3-sector tracks (e.g. sector 1 → sector 3 final, ~0.45 lap in ~1s) that were previously slipping through.

### Ghost icon / ghost formula

- Ghost icon and ghost comparison formula are now **only active when a ghost file was loaded from disk**. `ReferenceLapStore` tracks `fileLoadedClassIds` (populated only in `initialize()`) so in-session laps that get promoted to `persistedLaps` no longer falsely trigger the ghost UI.
- Beating/matching the ghost lap now shows **purple** (was green), consistent with the session-best colour.

23Kev:
Having a look at this one now, will need some testing and probably a review by @tariknz. I will ensure it does what it says on the box!

src/frontend/components/SectorDelta/SectorDelta.tsx
```diff
@@ -0,0 +1,190 @@
+import React, { useEffect } from 'react';
```
23Kev, line 1: I don't think we need to import React here. Check, but can be removed.

23Kev (commented):
One small unrequired import to clean up and will then approve.

dbadashboard:
## Latest changes (2026-04-21)

### Live sector delta via PCHIP interpolation

The current sector now shows a **live running delta** updated every tick, rather than a static comparison against the previous lap's sector time.

**How it works:**
- Uses `calculateReferenceDelta` from the Standings interpolation module (PCHIP hermite) to compute exactly how long the reference lap takes to reach the player's current track position from the sector entry point
- Compares that against the player's own elapsed time since entering the sector
- Result: a continuously updating delta that reflects real pace, not just the previous lap's number

**Reference selection:**
- Ghost file loaded + `prefer-ghost` setting → compares against the file-loaded ghost lap
- All other cases → compares against the player's **in-session personal best lap** (returns null until a personal best exists)

No fallback from personal best to ghost — the two modes are fully isolated.

### Refactor: removed duplicate interpolation code

`useLiveSectorDelta` previously called `interpolateAtPoint` directly (duplicating the Standings module). It now delegates to `calculateReferenceDelta` from `relativeGapHelpers`, which is the same function the Relative widget uses for gap calculations.

### Import cleanup

Removed the unused `import React` per review feedback — the `react-jsx` transform makes it unnecessary.

src/frontend/components/SectorDelta/SectorDelta.tsx
smithdt: removed, ty

smithdt:
Added a follow-up change for Sector Delta incident handling.

What changed:
- Off-track incidents now mark the current sector as unclean instead of dropping the sector time entirely.
- The Sector Delta widget shows the recorded delta with a small warning icon for unclean sectors.
- Truly invalid sectors from teleports, resets, or similar non-continuous movement are still ignored.

This means a spin or off-track that costs several seconds should now still show the real sector loss, while reset-style invalid sectors remain suppressed.

tariknz:
This looks great but can you review the CI error 
```

#
# Fatal error in , line 0
# Fatal JavaScript invalid size error 177765470 (see crbug.com/1201626)
#
#
#
#FailureMessage Object: 000000D65E3FE600
----- Native stack trace -----

 1: 00007FF7C8E93587 node::SetCppgcReference+21127
 2: 00007FF7C8D646CF node::TriggerNodeReport+71423
 3: 00007FF7CA063C15 V8_Fatal+197
 4: 00007FF7C9877E98 v8::Isolate::GetHeapProfiler+135912
 5: 00007FF7C951C8AC v8::base::AddressSpaceReservation::AddressSpaceReservation+216588
 6: 00007FF769A6DAFA 
```

I'd like to understand why this PR specifically tripped over the node memory limit?

smithdt:
## CI OOM fix

**Root cause:** `OverlayContainer.spec.tsx` (new in this PR) imports `OverlayContainer`, which statically imports `WIDGET_MAP` from `WidgetIndex`. That pulls in the full widget tree, including `TrackCanvas → tracks/tracks.json` (41 MB JSON file, 463 tracks). Parsing and instantiating that into live JS objects consumed ~3.1 GB of heap — exceeding the CI Node process limit and triggering the V8 fatal allocation error.

Existing specs avoid this: `TrackMap.spec.tsx` mocks `TrackCanvas` directly; `FlatTrackMap.spec.tsx` mocks `tracks.json` at the module level. But `OverlayContainer.spec.tsx` had no such mock.

**Fix:** Added `vi.mock('../../WidgetIndex', () => ({ WIDGET_MAP: {} }))` to the spec. The test only asserts that `useSectorTiming` is called — it never exercises widget rendering — so an empty map is sufficient.

**Before / after (single spec file, isolated run):**

| | Peak RSS | Import time |
|---|---|---|
| Before | 3.1 GB | 41.9 s |
| After | 250 MB | 0.97 s |

Pushed as a single follow-up commit on this branch.

smithdt:
@tariknz please check again - just needed to add a mock for the widget map for the test.

tariknz (approved):

PR #500, opened 2026-04-16
fix: driver tag Add New when groups are filtered

LanceVShoot:
# PR: Fix driver tag "Add New" when groups are filtered

## Description

Fixes the issue where clicking "Add New" in the Driver Tags settings ignored the active group filter and always defaulted to the first preset group. Now, when a group filter is applied, new driver tags are assigned to the filtered group by default.

Additionally, replaced the basic HTML `<select>` dropdown with a custom `GroupSelect` component that displays group icons and colors inline, providing better visual context when assigning tags to groups.

### How Tested

- [x] Tested in Driver Tags settings with group filters applied
- [x] Verified "Add New" defaults to filtered group when a filter is active
- [x] Verified "Add New" defaults to first group when no filter is active
- [x] Tested group selection dropdown displays icons and colors correctly

### Before
Basic HTML `<select>` dropdown with group names only, "Add New" always uses first group regardless of filter.

### After
Custom `GroupSelect` dropdown showing group icons/colors with inline preview, "Add New" respects active group filter.

Screenshot 2026-04-16 123359

Screenshot 2026-04-16 122518

## Type of Change

- [x] Refactoring (no functional changes)

## Checklist

### Modified
- `src/frontend/components/Settings/sections/DriverTagsSettings.tsx`
  - Added `activeGroupFilter` to `handleAddNew` dependency array
  - Use `activeGroupFilter` as default group when creating new entry
  - Replaced HTML `<select>` with custom `GroupSelect` component

### Created
- `src/frontend/components/Settings/sections/GroupSelect.tsx`
  - New custom dropdown component with group icon/color preview
  - Handles both 'tag' and 'badge' display styles
  - Click-outside detection to close dropdown

### DIFF
--- a/src/frontend/components/Settings/sections/DriverTagsSettings.tsx
+++ b/src/frontend/components/Settings/sections/DriverTagsSettings.tsx
@@ -13,6 +13,7 @@ import {
 import { PRESET_DRIVER_TAGS } from '../../../constants/driverTagBadges';
 import { renderDriverIcon } from '@irdashies/utils/driverIcons';
 import { IconPicker } from '../IconPicker';
+import { GroupSelect } from './GroupSelect';
 import { useDriverTagGlobalSettings } from './useDriverTagGlobalSettings';
 
 type IconMode = 'name' | 'image';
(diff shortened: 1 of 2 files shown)


23Kev (approved):
Will approve, but something to review another time. 
**Inline style objects for colour swatch**
The colour swatch <span> uses inline style for width, height, borderRadius, and background. Per project conventions (AGENTS.md: "Always use Tailwind — no custom CSS unless absolutely necessary") these should use Tailwind classes where possible. The dynamic background colour is the only one that genuinely needs inline style; the rest (w-3 h-3 rounded-sm flex-shrink-0) can be Tailwind.

PR #502, opened 2026-04-17
chore: allow more drivers to be displayed in standings

LanceVShoot:
## Description

Allow up to 63 (all) drivers to be displayed in standings in Standings Settings UI drop-down options.


23Kev (approved):
This is approved and will be merged.

PR #505, opened 2026-04-18
fix: session time during gridding

LanceVShoot:
During some sessions, the sessionTimeRemain and sessionTimeTotal from the SDK return back 604800 during the gridding process. This causes the Session Time in the sessionBar to show as 168 Hours. If in the grid phase, and the time remaining and time total are both 168h, then show nothing for the time as we don't have the information available.

## Before
Screenshot 2026-04-14 194054

## After
Screenshot 2026-04-18 105419


spikethedog78 (approved):
Simple and looks fine.

PR #511, opened 2026-04-21
fix/feat: "show only when on track" for lap timer

spikethedog78:
## Description

Added "show only when on track" option to the visibility settings for the Lap Timer

## Screenshots

image

## Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

- [X] I have discussed this change in the discord server
- [X] I have tested this in iRacing (either in an online session or with AI)

### DIFF
--- a/src/frontend/components/LapTimeLog/LapTimeLog.spec.tsx
+++ b/src/frontend/components/LapTimeLog/LapTimeLog.spec.tsx
@@ -38,6 +38,7 @@ const mockSettings = (
       practice: true,
       offlineTesting: true,
     },
+    showOnlyWhenOnTrack: true,
   };
 
   return {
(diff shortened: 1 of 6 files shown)


23Kev (commented):
A review showed up these two things. 
Request minor changes:

Gate current={isDriving ? data.current : undefined} on showOnlyWhenOnTrack as well, otherwise it's an unconditional behaviour change
Unify early returns to all use null instead of mixing null and <></>

spikethedog78:
> A review showed up these two things. Request minor changes:
> 
> Gate current={isDriving ? data.current : undefined} on showOnlyWhenOnTrack as well, otherwise it's an unconditional behaviour change Unify early returns to all use null instead of mixing null and <></>

I've replaced the <></> with null.

Not sure what you mean by gating that line also? It just hides the current lap timer when in the garage.

23Kev (approved):
Looks all good now

PR #515, opened 2026-04-23
fix: show irating change in prac/qual

LanceVShoot:
iRating change would only show in an official session, in the "Race" subsession. This update places the check on the EventType instead of the sessionType, checking for "Race" EventType. iRating Change will show in practice/qualifying/race, as long as it's an official race session. It will not show for practice sessions outside of the official lobby.

Before/After Practice
Screenshot 2026-04-23 030314

Screenshot 2026-04-23 030245

Before/After Qualifying
Screenshot 2026-04-23 030422

Screenshot 2026-04-23 030436

After Qualifying and positions changed
Screenshot 2026-04-23 030515

Race remains unchanged
Screenshot 2026-04-23 030540

### DIFF
--- a/src/frontend/components/Standings/hooks/useDriverStandings.tsx
+++ b/src/frontend/components/Standings/hooks/useDriverStandings.tsx
@@ -15,6 +15,7 @@ import {
   usePitLap,
   usePrevCarTrackSurface,
   useLapTimeHistory,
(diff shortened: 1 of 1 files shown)


spikethedog78 (approved):
Looks good.

PR #522, opened 2026-04-25
fix(standings): position bug + perf(sectordelta): drop 60Hz re-renders

tariknz:
## Summary

Two independent fixes from an audit of changes since v0.2.1.

### 1. `fix:` standings position for not-yet-in-results drivers in multi-class

Introduced in #518. `notYetInResults` (drivers without lap times) assigned the global `position` field as `mapped.length + classPositionCounts.size`, where `.size` is the number of distinct class IDs in the map — not an incrementing counter. Every such driver got the same `position`, breaking ordering after the position-based sort in `useDriverStandings`. Switched to `mapped.length + index + 1`.

### 2. `perf:` round `LapDistPct` / `SessionTime` in SectorDelta to drop 60Hz re-renders

Introduced in #488. `SectorDelta` and `useLiveSectorDelta` subscribed to `useTelemetryValue('LapDistPct')` and `'SessionTime'` — both unrounded scalar floats updating every tick — so the component re-rendered and the PCHIP-interpolation `useMemo` re-ran 60×/sec.

Added `useTelemetryValueRounded` as a scalar mirror of the existing `useTelemetryValuesRounded` array hook and used it in the SectorDelta consumers. `useSectorTiming` is intentionally left unrounded — it interpolates between consecutive frames to detect sector boundary crossings and needs the full signal (same reasoning as `CarSpeedsStore` per `AGENTS.md`).

Precision choices follow `AGENTS.md`:
- `LapDistPct`, 3dp — progress bar (~5m on a 5km track, sub-pixel)
- `LapDistPct`, 4dp — `useLiveSectorDelta`, matches `REFERENCE_INTERVAL = 0.0025`
- `SessionTime`, 2dp — 10ms; display rounds to 0.1s anyway

## Test plan

- [x] `npm run test` (744 passing)
- [x] `npm run lint` (clean, including `tsc --noEmit`)
- [ ] Multi-class race with drivers who haven't yet set a lap — confirm they appear at the bottom in stable car-number order
- [ ] Single-class warmup — confirm ordering unchanged
- [ ] On-track with SectorDelta widget visible — confirm sector progress bar still animates smoothly and live delta updates
- [ ] React DevTools profiler on SectorDelta — confirm renders/sec dropped from ~60 to a handful

### DIFF
--- a/src/frontend/components/SectorDelta/SectorDelta.tsx
+++ b/src/frontend/components/SectorDelta/SectorDelta.tsx
@@ -1,7 +1,11 @@
 import { useEffect } from 'react';
 import { GhostIcon, WarningIcon } from '@phosphor-icons/react';
 import { useSectorDeltas, useSectorTimingStore } from '@irdashies/context';
-import { useSessionVisibility, useTelemetryValue } from '@irdashies/context';
+import {
+  useSessionVisibility,
+  useTelemetryValue,
+  useTelemetryValueRounded,
+} from '@irdashies/context';
 import { useReferenceLapSectorTimes } from '@irdashies/context';
 import type { SectorDeltaConfig } from '@irdashies/types';
 import type { TimeFormat } from '@irdashies/types';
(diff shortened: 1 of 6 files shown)


tariknz:
@dbadashboard can you take a look at this? is it ok if we round the telemetry values? These have a very high precision by default. This will be the same rounding as the trackmap currently.

smithdt:
@tariknz Oh yeah, rounding to these should be completely fine and not even noticeable.

PR #569, opened 2026-05-28
feat: add heart rate widget (HypeRate)

ziadkadry99:
## Description

Adds a **Heart Rate** widget that displays your live heart rate from [HypeRate](https://www.hyperate.io/) — no API key or account required. You enter your session id (the code from your HypeRate share link, e.g. `app.hyperate.io/`**`KiY`**) and the widget embeds HypeRate's overlay.

- **No key/account** — just the session id
- **Optional widget/theme** — paste any HypeRate widget link or just its name (e.g. `Bouncing_Heart_Widget`), or an `app.hyperate.io/animation/<n>/...` route. A forgiving URL builder normalizes all of these and always substitutes your session id (any id baked into the pasted URL is ignored)
- **Transparent + no scrollbar** — the embedded page is restyled the way OBS does it (injected `background: transparent` CSS)
- **Auto-fit** — measures the graphic inside the page and scales/centres it to the widget, so the whole graphic shows and scales on resize (no truncation)
- Visible even when iRacing isn't running (like the Twitch Chat widget)

Heart-rate data is independent of iRacing telemetry; the screenshot below shows the overlay running over a live iRacing session.

Implementation notes:
- The embed renders in a `<webview>` (falls back to an `<iframe>` in Storybook / browser source) so transparency CSS can be injected cross-origin. This required enabling `webviewTag` on the overlay window; the guest keeps Electron's secure defaults (no node integration, context isolation on).
- URL building is a pure, unit-tested helper (`hyperateUrl.ts`).

## Screenshots
Screenshot 2026-05-28 000515
Screenshot 2026-05-28 002153

### Before

N/A — new widget.

### After

HeartRate overlay (HypeRate) shown top-centre over a live iRacing session.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added Heart Rate overlay widget (HypeRate) embeddable via Session ID; supports optional custom widget/theme URLs, auto-scales to widget size, and remains visible even when iRacing is not running
  * Settings UI entry and default dashboard widget added; settings let you enter Session ID and optional widget/theme URL
  * Storybook previews for the widget and flexible widget URL/session-id handling for common input forms
  * OBS/browser-source friendly (transparent background, no scrollbars)

* **Documentation**
  * README updated with Heart Rate overlay setup instructions and screenshot

[Review Change Stack](https://app.coderabbit.ai/change-stack/tariknz/irdashies/pull/569?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


ziadkadry99:
Thanks for the review! Addressed in `3b33acd`:

- **Restrict embeds to HypeRate hosts** — `buildHyperateEmbedUrl` now falls back to the default overlay for any non-`hyperate.io` host, and the overlay window adds a `will-attach-webview` guard that blocks non-HypeRate `src` and forces secure guest prefs (no preload, `nodeIntegration: false`, `contextIsolation: true`).
- **Deep-merge saved config** — switched to `deepMergeConfig` so a partial `sessionVisibility` no longer drops the other default flags.
- **Type-guard the webview result** — `executeJavaScript` output is validated with an `isBox` guard instead of a cast.
- **Trim `deviceId`** in the configured-state check.
- **Added `TelemetryDecorator()`** to the Storybook stories.

One I left as-is: wrapping the widget root in `memo()`. `WIDGET_MAP` is typed `(config?) => Element | null`, and a `MemoExoticComponent` isn't assignable to that — all the other registered widget roots (SlowCarAhead, Weather, TwitchChat, …) are plain function components, so I kept it consistent and added a comment noting why.

ziadkadry99:
Addressed the follow-up in `a764c77`: on `did-attach-webview` the guest now denies window opens (`setWindowOpenHandler` → deny) and prevents navigation (`will-navigate` → preventDefault), so an embedded HypeRate page can't open popups or navigate away after attachment — defense in depth alongside the `will-attach-webview` host/prefs check.

tariknz:
> Addressed the follow-up in `a764c77`: on `did-attach-webview` the guest now denies window opens (`setWindowOpenHandler` → deny) and prevents navigation (`will-navigate` → preventDefault), so an embedded HypeRate page can't open popups or navigate away after attachment — defense in depth alongside the `will-attach-webview` host/prefs check.

Hey are you on our discord?

ziadkadry99:
@tariknz Yes I am! @thesnake91

tariknz (commented):
Nice work on the hardening here — the `will-attach-webview` host allowlist, secure guest defaults, popup denial and `will-navigate` containment are all correct, and the `insertCSS`/`executeJavaScript` calls only run constant strings in the guest's isolated context. A few defense-in-depth items I'd like to see before merge.

### 1. Enable the full sandbox on the guest (should-fix)
`will-attach-webview` sets `contextIsolation`/`nodeIntegration` but not `sandbox`. Since the guest renders untrusted third-party content, turning on the full sandbox is a cheap, strong addition:

```js
browserWindow.webContents.on('will-attach-webview', (event, webPreferences, params) => {
  delete webPreferences.preload;
  webPreferences.nodeIntegration = false;
  webPreferences.contextIsolation = true;
  webPreferences.sandbox = true; // add
  ...
});
```

### 2. Isolate the guest's session with a partition (should-fix)
The `<webview>` has no `partition`, so HypeRate shares the **default session** (cookies/cache/storage) with the app's own renderer content. One attribute walls it off:

```js
webview.setAttribute('partition', 'persist:hyperate'); // or in-memory: 'hyperate'
```

### 3. Contain cross-origin redirects, not just navigations (should-fix)
The guest is pinned via `will-navigate`, but HTTP 3xx redirects fire `will-redirect`, **not** `will-navigate`. So after the allowlisted initial load, the top frame could be redirected off `hyperate.io` without being blocked. Blast radius is limited (sandboxed, no node), but it's a gap in the stated "guest can only ever be HypeRate" model:

```js
guest.on('will-redirect', (e, url) => {
  let host = '';
  try { host = new URL(url).hostname; } catch { /* blocked below */ }
  if (!/(^|\.)hyperate\.io$/i.test(host)) e.preventDefault();
});
```

Ideally factor the `/(^|\.)hyperate\.io$/i` check into a shared helper so the `will-attach-webview` and `will-redirect` copies can't drift apart.

---

### Non-blocking context

- **`webviewTag: true` is enabled on every overlay window**, not just one hosting this widget. Because `webPreferences` are fixed at window-creation and any widget can land on any display, scoping it cleanly would mean recreating windows on settings toggle — not worth the complexity given the allowlist already constrains attachable guests to `hyperate.io`. Suggest just a one-line comment at the `webviewTag: true` site noting it's global-by-design and constrained by the allowlist below, so it reads as a documented decision rather than an oversight.
- This is the first time the app embeds arbitrary third-party web content in Electron (the TwitchChat widget renders natively via API, it doesn't embed). With the items above, the guest's blast radius ≈ a normal sandboxed browser tab, which is fine — just flagging the new third-party trust relationship. Defaulting the widget to `enabled: false` is the right call.

tariknz:
Sorry just added some claude suggestions have a look - just concerned about thirdparty integrations but I appreciate already the mitigations you put in!

Otherwise I'm happy with it to go in

ziadkadry99:
Thanks for the careful review @tariknz — all three should-fix items addressed in d4181b1, plus the non-blocking comment ask and a CI fix from the main merge.

**1. Full sandbox on the guest** — `webPreferences.sandbox = true` in `will-attach-webview` alongside the existing `contextIsolation`/no-`nodeIntegration`/no-`preload`.

**2. Isolated session partition** — `<webview>` now sets `partition="persist:hyperate"` in `HeartRateEmbed.tsx`, so HypeRate's cookies/cache/storage are walled off from the default app session.

**3. Cross-origin redirect containment + shared host helper** — factored `/(^|\.)hyperate\.io$/i` into `isAllowedGuestHost(url)` at the top of `overlayManager.ts`, and added a `will-redirect` handler that calls `preventDefault()` if a 3xx tries to redirect off `hyperate.io`. `will-navigate` was tightened to use the same allowlist (was previously a blanket block — same effect for the pinned-host case, but symmetric with `will-redirect`). Both `will-attach-webview` and the post-attach handlers now share the single check.

**Non-blocking comment** — expanded the comment at `webviewTag: true` to call out that it's global-by-design on every overlay window (since `webPreferences` are fixed at window-creation and any widget can land on any display), constrained by the allowlist below, and pointing to the guest's full hardening set.

**CI fix** — the merge of `main` dropped `sessionVisibility` from the heartrate default config, which broke `tsc`. Restored.

Lint + heart-rate unit tests green locally. Let me know if anything else needs tightening.

tariknz:
woops sorry for breaking the config

tariknz (approved):

PR #584, opened 2026-06-11
feat: Race Distance based by Laps and Time + Multiclass sorting

L061N:
## Description

I combined the useTotalRaceLaps and useTotalRaceTime into one file to cover the odd case of a session having X laps and is limited by time. Example: 35 lap race 15min, with 1 minute laps only 15 laps could be done.

Also, changed the sorting of class speeds, first is by the average position of the top 2 cars in class then by relative speed as a fall back. Covers cases similar to TCR class being fast than GT4, may want to tweak priority so p1 overall is the fastest class.
Should note iracing has track of the top 8 classes and colors.

## Type of Change

- [/] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

- [ ] I have performed a self-review of my own code

## Summary by CodeRabbit

* **Improvements**
  * Better class standings ordering that prioritizes top-car timing and positions.
  * Consolidated race timer and lap calculations for more consistent lap-overrun handling (race sessions only).
  * More accurate lap counting and race time estimates for both fixed-lap and time-limited formats.
  * Fuel estimates now use the consolidated total-race values for more consistent results.


src/frontend/components/Standings/createStandings.ts
```diff
@@ -446,7 +446,57 @@ export const groupStandingsByClass = (
...
+              return {
+                  average:
+                      positions.reduce((sum, pos) => sum + pos, 0) /
+                      positions.length,
+                  bestPosition: positions[0], // lower is better
+              };
+          };
+
+          const statsA = getTop2Stats(classA);
+          const statsB = getTop2Stats(classB);
```
tariknz: **Perf:** this comparator runs in a hot path. `groupStandingsByClass` is called inside the `useDriverStandings` memo, whose deps include `carIdxLapDistPct`/`carIdxLap` — so it recomputes many times per second during a race.

The comparator went from O(1) (`relativeSpeed` subtraction) to O(n log n) + 3 array allocations (`filter`/`sort`/`slice`) **per comparison**, and `getTop2Stats(classX)` is recomputed `log C` times for the same class instead of once.

Absolute cost is small since iRacing caps classes at 8, so it won't cause visible jank — but it's easy to remove the GC churn and redundant work by hoisting the stats out of the comparator so each class is computed once:

```ts
const statsByClass = new Map(
  sorted.map(([id, drivers]) => [id, getTop2Stats(drivers)])
);
sorted.sort(([idA, classA], [idB, classB]) => {
  const statsA = statsByClass.get(idA);
  const statsB = statsByClass.get(idB);
  // ...same comparison logic, no allocations in the comparator
});
```

Behavior stays identical; this just moves the work to O(C) once-per-class instead of O(C log C) inside the sort.

tariknz (approved):

PR #600, opened 2026-06-20
feat: dynamic per-widget show/hide hotkeys

tariknz:
## Description

Adds a **dynamic, per-widget show/hide hotkey** so each widget can be toggled on its own. Bindings are **unbound by default** and are derived live from the current dashboard's widget list, so newly added widgets automatically get a bindable toggle with no per-widget code.

Implementation reuses the existing keybinding pipeline (keyboard **and** gamepad via WebHID) and mirrors the existing global `Alt+H` transient-hide pattern:

- A widget toggle is the dynamic action id `toggle-widget:<widgetId>` (per widget instance). These have no factory default and are persisted only once the user binds one.
- Pressing the key flips a **session-only** hidden-set in the main process (`KeybindingManager.toggleWidgetHide`) and broadcasts `widget-toggle-hide`; `OverlayContainer` filters that widget out. The saved `enabled` config is **never** mutated, so hidden state resets on restart — exactly like the global hide.
- Settings → **Key Bindings** now has a **Widget Visibility** section listing every widget in the current dashboard, each with a recorder row (shows "Not set" until bound; reset/clear unbinds).

Storage/types changes: `KeybindingActionId` split into `StaticKeybindingActionId | WidgetToggleActionId`; `getKeybindings`/`updateKeybinding`/`resetKeybinding` handle dynamic create/unbind and ignore empty accelerators in conflict checks; shared helpers added in `@irdashies/shared`.

Known limitation: a binding for a widget later removed from the dashboard becomes an orphaned stored entry — its shortcut registers but toggles nothing and it won't show in settings (no-op). Left for a later cleanup.

Tested via unit tests (shared helpers + storage dynamic behavior) and the full suite (958 passing); not yet verified in a live iRacing session.

### Before
Key Bindings settings showed only the three fixed shortcuts (Hide/Show UI, Edit Layout, Save Telemetry). No way to show/hide an individual widget.

### After
Key Bindings settings adds a **Widget Visibility** section with one bindable (keyboard or controller) show/hide toggle per widget, unbound by default.

image

_(Storybook story added: `components/Settings/KeybindingsSettings`.)_

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added hotkey support for showing/hiding individual dashboard widgets, with widget visibility bindings now managed from the keybinding settings.
  * Keybinding recording/display was enhanced to carry additional label/description details for widget-toggle bindings.
* **Bug Fixes**
  * Unbinding now properly removes widget visibility bindings when a hotkey is cleared.
* **Tests**
  * Expanded unit and storage test coverage for widget-toggle actions, including update/reset behavior and default/override merging.


tariknz:
Thanks for the review @coderabbitai — addressed in `0dfedc82`:

- **Validate `actionId` at the IPC boundary** ✅ Added `assertSupportedAction` in `keybindingsBridge.ts` — `keybindings:update` now rejects anything that isn't a known static action or a well-formed `toggle-widget:<id>` (non-empty id) before it reaches storage.
- **Reject malformed `toggle-widget:` keys on merge** ✅ `getKeybindings()` now uses the new `isValidWidgetToggleActionId` shared helper, so empty-id keys are dropped.
- **Nitpick: meta-forwarding test** ✅ Added (plus an unsupported-action rejection test).
- **Nitpick: `TelemetryDecorator()` in story** ✅ Added.

**Intentionally not doing — replay hidden state to late/reloaded overlays:** this matches the existing global `Alt+H` hide, which is also broadcast-only and never replayed to windows created later. The per-widget hide is session-only/transient by the same design, so adding a sync path would make it inconsistent with the established pattern. If we ever want hide state to survive window recreation, it'd be a follow-up that covers both toggles together.

Added `isValidWidgetToggleActionId` + tests; full suite green (962 passing).

PR #607, opened 2026-06-23
fix: Class Sorting Bug

L061N:
## Description

Previously used the average overall position of the top 2 cars in each class to sort classes. 
Bug: In small field sizes if p1 in class A is ahead of class B, but p2 in class A is behind 3+ cars in class B, class B would be considered faster than class A.
Fix: just use the overall position of the leader in each class, then sort by iracing's class relative speed. 

Bit of an odd visual, in qualifying if class B is faster than class A the order is Class B -> Class A. If grid by class is enabled iracing can switch the order to Class A -> Class B at the race start until p1 in Class B passes P1 in Class A.

## Type of Change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

- [ x] I have tested this in iRacing (either in an online session or with AI)
- [ x] All tests pass locally via `npm test`
- [ ] I have performed a self-review of my own code

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated standings calculations to prioritize the single class leader’s position rather than averaging across the top multiple drivers.

### DIFF
--- a/src/frontend/components/Standings/createStandings.ts
+++ b/src/frontend/components/Standings/createStandings.ts
@@ -447,7 +447,7 @@ export const groupStandingsByClass = (
         (d) =>
           d.position !== undefined &&
           d.fastestTime !== undefined &&
-          d.fastestTime > 0
+          d.fastestTime > 0 
       )
       .sort((a, b) => (a.position ?? 999) - (b.position ?? 999))
       .slice(0, 1); // Changed .slice(0,2) to .slice(0,1) to only compare leader of each class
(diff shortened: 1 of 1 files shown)


L061N:
This has a quick fix to edge cases of classes with one car being ahead of another class but p2 is behind 2 cars of the other class.
I'm working on adding a setting to make this optional. [ClassSortingSettings](https://github.com/L061N/irdashies/tree/ClassSortSettings)

tariknz (approved):

PR #640, opened 2026-07-27
feat: add session bar items - fuel, laps, top speed, manufacturer position, class rank

Foribald:
## Description

Adds new configurable items to the standings and relative widget header/footer session bars:

- **Fuel Level** – current fuel in litres or gallons
- **Last Lap** – player's last lap time
- **Best Lap** – player's best lap time, session best shown in green
- **Top Speed** – last lap top speed and session best
- **Manufacturer Position** – manufacturer icon + player's rank among same-make drivers (e.g. Dallara icon + 2/4), with optional hide when single make in session or when player is the only driver of their make
- **Class Rank** – racing helmet icon + player's class position out of total class cars (e.g. 4/11)

All items are disabled by default and independently configurable per bar (header/footer) for both standings and relative widgets.

Also fixes demo mode car ID (97 → 99, Dallara IR18) so the manufacturer icon renders correctly in Storybook and demo mode.

## Screenshots
new-header-footer-standings-relative - config
new-header-footer-standings-relative

### Before
No manufacturer position or class rank items in the session bar.

### After

Settings panel — Manufacturer Position enabled with sub-options, Class Rank enabled:

Race view — header bar showing manufacturer position (d 2/4) and class rank (4/11):

## Type of Change

## Checklist

## New Features
* Expanded the Session Bar with configurable fuel level, lap times, top speed, manufacturer position, and class rank metrics.
* Added live last-lap and session-best top-speed displays with metric/imperial unit support.
* Added options to hide manufacturer position for single-make or single-driver sessions.
* Updated default dashboard layouts to include the new metrics.
* Added a racing helmet icon for class-rank displays.

## Bug Fixes
* Improved demo-mode driver data mapping for more accurate displayed information.


Foribald:
screenshots
new-header-footer-standings-relative - config
new-header-footer-standings-relative

src/frontend/components/Standings/components/SessionBar/SessionBar.tsx
```diff
@@ -151,6 +165,20 @@ export const SessionBar = ({
...
+  const lastLapTopSpeedMs = useLastLapTopSpeed();
+  const sessionBestTopSpeedMs = useSessionBestTopSpeed();
+  const drivers = useSessionDrivers();
+  const playerCarIdx = useDriverCarIdx();
+  const carIdxPositions = useTelemetryValues('CarIdxPosition');
+  const carIdxClassPositions = useTelemetryValues('CarIdxClassPosition');
+  const sessionBestLap =
+    allBestLapTimes && allBestLapTimes.some((t) => t > 0)
+      ? Math.min(...allBestLapTimes.filter((t) => t > 0))
+      : undefined;
```
tariknz: I'm concerned here suddenly SessionBar has all these dependencies - these should all go in their own components instead of just implemented in the render function. I suspect there's a bit performance hit here.

Foribald: on my tests I didn't see bad performance but maybe it depends on number of cars.
I am asking Claude to refactor it

Foribald:
Addressed the review comment about `SessionBar` having too many dependencies handled directly in the render function, in two commits:

**`4fe8cd3` — split SessionBar into per-item components**
Each of the 19 configurable items (session time, laps, fuel, brake bias, manufacturer position, etc.) now lives in its own memoized component that only subscribes to the telemetry/session data it actually needs, instead of one component unconditionally calling ~19 hooks and building a giant inline render-closure map regardless of which items are enabled. Null-capable items return `null` and self-filter; first/last alignment (for the standalone `InformationBar` layout) moved from index bookkeeping in the parent to CSS structural selectors (`first:`/`last:`), since the parent can no longer know ahead of time which items will render `null`.

**`5d36cce` — dedupe the timing computation this introduced**
Splitting `sessionTime` and `sessionLaps` into separate components meant each one independently called `useSessionLapCount()` *and* `useTotalRaceValue()` (which itself calls `useSessionLapCount()` again) — a leader-car-finding loop, average-lap-time math, and a `useEffect` with ref-tracked green-flag/checkered-lap state, running 4x per render instead of once whenever both items are enabled (the default for header+footer). Fixed by following the existing `TopSpeedStoreUpdater` precedent in this codebase: compute once in a dedicated non-rendering `SessionTimingStoreUpdater`, write to a small Zustand store, have the two items read from it instead of calling the hooks directly. Hook internals are untouched — zero behavior risk, just resourced from one call site instead of four.

(Also caught along the way: `SessionBar.stories.tsx` renders `<SessionBar>` standalone rather than through `Standings`/`Relative`/`InformationBar`, so it needed the new updater mounted in its decorator too, or the `sessionTime`/`sessionLaps` stories would've shown frozen zero values.)

**Scope note:** investigation surfaced that `TitleBar` also independently calls `useSessionLapCount()` plus a third near-duplicate hook (`useTotalRaceLaps`, with a *different* fixed-lap-race formula than `useTotalRaceValue`). Left that out of scope here — it's pre-existing tech debt unrelated to this PR's `SessionBar` change and would need its own careful investigation before touching.

**Verification:** `tsc --noEmit` clean, `eslint .` clean (project-wide), full suite passing (94 files / 1085 tests, including new specs for the split components and the new store/updater). Manually checked the Storybook stories for `SessionBar` render identically to before.

src/frontend/components/Standings/components/SessionBar/components/ManufacturerPositionItem/ManufacturerPositionItem.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { CarManufacturer } from '../../../CarManufacturer/CarManufacturer';
+import { CAR_ID_TO_CAR_MANUFACTURER } from '../../../CarManufacturer/carManufacturerMapping';
+import { sessionBarItemWrapperClass } from '../../sessionBarItemWrapperClass';
+import type { SessionBarItemProps } from '../../sessionBarItemTypes';
+
+export const ManufacturerPositionItem = memo(
+  ({ settings, standalone }: SessionBarItemProps) => {
+    const drivers = useSessionDrivers();
+    const playerCarIdx = useDriverCarIdx();
+    const carIdxPositions = useTelemetryValues('CarIdxPosition');
```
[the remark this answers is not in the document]
Foribald: the suggestion is wrong

Two independent problems with it:

useTelemetryValuesRounded doesn't round the returned values. Looking at [TelemetryStore.tsx:65-73](vscode-webview://1e4sjfuc3000o27bft7m4ko7o9esh0scdkpestlkvce4uih4pfgi/src/frontend/context/TelemetryStore/TelemetryStore.tsx#L65-L73), the selector still returns the raw, unrounded array — arrayCompareRounded is only used as the equality comparator to decide whether to skip a re-render. The values ManufacturerPositionItem sorts by would be identical either way.

CarIdxPosition has no sub-resolution component to filter. It's an integer race position (1st, 2nd, 3rd...) — [_GENERATED_telemetry.ts:95](vscode-webview://1e4sjfuc3000o27bft7m4ko7o9esh0scdkpestlkvce4uih4pfgi/src/app/irsdk/types/_GENERATED_telemetry.ts#L95) types it as number[], and iRacing reports it as a whole number, unlike CarIdxLapDistPct (a genuine continuous 0.0–1.0 float). Rounding an already-integer value to 3 decimal places is a no-op — arrayCompareRounded(3, a, b) and exact arrayCompare(a, b) produce identical results here, since any actual change is a whole-number change (an overtake) that should trigger a re-render anyway.

The AGENTS.md guideline CodeRabbit cited is about a different field. I checked every existing use of useTelemetryValuesRounded('CarIdx...', 3) in this codebase (18 call sites) — every single one operates on CarIdxLapDistPct (or CarIdxEstTime/CarIdxF2Time), never CarIdxPosition. More tellingly, the actual "position sorting (standings)" code this guideline describes — [useDriverPositions.tsx:45](vscode-webview://1e4sjfuc3000o27bft7m4ko7o9esh0scdkpestlkvce4uih4pfgi/src/frontend/components/Standings/hooks/useDriverPositions.tsx#L45) and useDriverStandings.tsx — reads CarIdxPosition via plain useTelemetry/useTelemetryValues, not the rounded variant. CodeRabbit pattern-matched on the word "position" and applied a rule meant for CarIdxLapDistPct to a structurally different field.

Foribald:
@coderabbitai Fixed in \`7c46da8\` — confirmed this was a real, pre-existing gap (not introduced by this PR): \`TopSpeedStoreUpdater\` was only mounted in \`Standings.tsx\`/\`Relative.tsx\`, so \`InformationBar.tsx\` (which exposes \`topSpeed\` via \`InformationBarConfig extends SessionBarConfig\`) and the \`SessionBar\` Storybook decorator would silently show the \`—\` placeholder instead of real speed data if \`topSpeed\` were enabled there.

Mounted \`<TopSpeedStoreUpdater enabled={!!settings?.topSpeed?.enabled} />\` in \`InformationBar.tsx\` (same gating pattern as the other two call sites), and unconditionally in the \`SessionBar.stories.tsx\` decorator (mirroring how \`SessionTimingStoreUpdater\` is already handled there). Verified with `tsc --noEmit`, `eslint .`, and the full test suite (94 files / 1085 tests) — all clean.

src/frontend/components/Standings/Standings.tsx
```diff
@@ -90,6 +94,15 @@ export const Standings = () => {
...
     >
+      <TopSpeedStoreUpdater
+        enabled={
+          !!(
+            settings?.headerBar?.topSpeed?.enabled ||
+            settings?.footerBar?.topSpeed?.enabled
+          )
+        }
+      />
+      <SessionTimingStoreUpdater />
```
tariknz: Once you mount TopSpeedStoreUpdater you will still run every hook inside of it regardless of it returning conditionally based on the enabled state.

Second issue is actually both of these updaters are on the wrong level. If you are displaying Relative, Standings, and InformationBar in the same screen all these will update the store simultaneously.

tariknz: I'd feel better if you remove TopSpeed and SessionTiming (less concerning) features into a seperate PR

Foribald:
@coderabbitai Fixed in \`d2e7903\` — confirmed this was accurate. \`useSessionDrivers()\` returns the raw, unfiltered driver list (pace cars/spectators included), while the rest of the standings path (\`createStandings.ts\`, \`useCarClassStats.tsx\`, \`useIsSingleMake.ts\`) consistently filters \`CarIsPaceCar\`/\`IsSpectator\` before counting. \`ClassRankItem\` didn't, so a pace car or spectator sharing the player's \`CarClassID\` inflated the total (e.g. \`2/4\` instead of \`2/3\`).

Applied the same filter to both \`ClassRankItem.tsx\` and \`ManufacturerPositionItem.tsx\` (which had the identical pattern for manufacturer counts, not just class counts). Both were carried over verbatim from the original monolithic \`SessionBar.tsx\`, so this predates the item-split refactor.

Added regression specs for both components covering the pace-car/spectator exclusion scenario specifically. Verified with \`tsc --noEmit\`, \`eslint .\`, and the full suite (96 files / 1087 tests) — all clean.

Foribald:
@coderabbitai Fixed in \`b8d9cb7\` — both points confirmed accurate.

**Wrong level / simultaneous updates:** \`TopSpeedStoreUpdater\` and \`SessionTimingStoreUpdater\` were mounted once per widget (Standings, Relative ×2 branches, InformationBar). With Standings + Relative + InformationBar all on screen at once, that's up to 4 instances independently writing to the same global store per render. Moved both to a single mount in \`OverlayContainer.tsx\` — the actual common ancestor of every widget on a display — matching the existing \`SectorTimingUpdater\`/\`PushToPassUpdater\`/\`PitLapUpdater\` pattern already used there for the same reason.

**Hooks running regardless of enabled:** confirmed — passing \`enabled\` as a *prop* into an always-mounted component doesn't stop React from calling the hooks inside it; only the store *write* was gated, not the \`useTelemetryValue('Speed')\` etc. subscriptions themselves. Fixed by having the new \`OverlayContainer\`-level wrappers (\`TopSpeedUpdater\`, \`SessionTimingUpdater\`) conditionally *mount* the store updater (\`if (!enabled) return null\`) instead of always mounting it with \`enabled={false}\` — so when nothing needs the data, the component genuinely isn't in the tree and its hooks never run. Verified with specs asserting the inner updater is `not.toHaveBeenCalled()` in the disabled case.

While auditing for the same pattern, found two more pre-existing duplicated hooks in older SessionBar items and applied the same treatment: \`useTrackTemperature\` (\`AirTemperatureItem\`/\`TrackTemperatureItem\` each ran their own \`TrackTempCrew\`/\`AirTemp\` telemetry subscriptions) and \`useSessionBestLapTime\` (\`LastLapItem\`/\`BestLapItem\` each ran their own \`CarIdxBestLapTime\` filter+min). \`TrackTemperatureStore\` holds raw Celsius rather than pre-formatted strings, since sibling widgets can have different unit settings — formatting was pulled out into a pure \`formatTemperature()\` helper each item calls with its own unit preference, preserving correct per-widget behavior while still deduping the actual telemetry subscription.

Verified with \`tsc --noEmit\`, \`eslint .\`, and the full suite (105 files / 1114 tests, 19 new) — all clean.

src/frontend/components/Standings/Standings.tsx
Foribald: are you sure? in the last commit I resolve the 2 issues you mentionned. Furthermore I optimized Tracktemp as useTrackTemperature() was running once each in AirTemperatureItem and TrackTemperatureItem.
Please look at the last commit and advise

tariknz: No I think this looks much better. I'll check again tonight. Thank you!

tariknz (approved):

PR #641, opened 2026-07-27
feat: add compact SoF and per-manufacturer car counts to standings class header

Foribald:
## Description

Enhances the standings widget class header row with two new display options:

**Compact SoF formatting** (optional toggle)
- Displays Strength of Field in a compact format for values >= 1000 (e.g. `2.4k` instead of `2400`)
- Controlled per class header via a settings toggle

**Per-manufacturer car counts** (optional, configurable)
- Shows manufacturer icon + driver count for each make within the class (e.g. Toyota icon + `7`)
- Manufacturers sorted by count descending (most represented first)
- Configurable cap from 1 to 10 or All — a `+N` overflow badge appears when manufacturers are hidden
- Optional "always show player manufacturer" override: if the player make falls outside the cap, it replaces the last visible slot so the driver always sees their own count
- Hidden automatically when all cars in the class share the same manufacturer

## Screenshots
standings compact SoF and per-manufacturer car counts -config
standings compact SoF and per-manufacturer car counts

### Before
Class header shows class name, SoF and total driver count only.

## Type of Change

## Checklist

## New Features

* Added optional compact Strength of Field formatting in standings headers.
* Added manufacturer count displays for each class, with configurable limits.
* Added an option to always show the player’s manufacturer.
* Added settings controls to enable and customize manufacturer statistics.
* Improved single-make detection based on standings display preferences.

### DIFF
--- a/src/frontend/components/Standings/Standings.tsx
+++ b/src/frontend/components/Standings/Standings.tsx
@@ -26,6 +26,7 @@ import {
 } from '@irdashies/context';
 import { useIsSingleMake } from './hooks/useIsSingleMake';
 import { computeStintLap } from './components/DriverInfoRow/cells/lapCountUtils';
+import { CAR_ID_TO_CAR_MANUFACTURER } from './components/CarManufacturer/carManufacturerMapping';
 
 export const Standings = () => {
   const settings = useStandingsSettings();
@@ -120,6 +121,18 @@ export const Standings = () => {
                   ? (classColorHex ?? highlightHex)
                   : highlightHex;
 
+            const mfrMap = new Map<string, { carId: number; count: number }>();
+            for (const s of classStandings) {
+              if (s.carId === undefined) continue;
+              const mfr =
+                CAR_ID_TO_CAR_MANUFACTURER[s.carId]?.manufacturer ?? 'unknown';
+              if (!mfrMap.has(mfr)) mfrMap.set(mfr, { carId: s.carId, count: 0 });
+              mfrMap.get(mfr)!.count += 1;
+            }
+            const manufacturerCounts = Array.from(mfrMap.values()).sort(
+              (a, b) => b.count - a.count
+            );
+
             return classStandings.length > 0 ? (
               <Fragment key={classId}>
                 <DriverClassHeader
@@ -135,6 +148,7 @@ export const Standings = () => {
                   colSpan={100}
                   classHeaderStyle={settings?.classHeaderStyle}
                   compactMode={generalSettings?.compactMode}
+                  manufacturerCounts={manufacturerCounts}
                 />
                 {classStandings.map((result, driverIndex) => {
                   const prev = classStandings[driverIndex - 1];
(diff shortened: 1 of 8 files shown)


src/frontend/components/Standings/Standings.tsx
```diff
@@ -120,6 +121,30 @@ export const Standings = () => {
...
+                ? (CAR_ID_TO_CAR_MANUFACTURER[playerCarId]?.manufacturer ?? 'unknown')
+                : undefined;
+            const playerManufacturerEntry = playerMfr
+              ? manufacturerCounts.find(
+                  (m) =>
+                    (CAR_ID_TO_CAR_MANUFACTURER[m.carId]?.manufacturer ?? 'unknown') ===
+                    playerMfr
+                )
+              : undefined;
+
```
tariknz, line 154: Skip aggregation unless manufacturerStats.enabled is true. Ideally, memoize it using stable driver/class/car identity data

useMemo([standings]) alone will still recalculate whenever telemetry recreates standings.

src/frontend/components/Standings/Standings.tsx
```diff
@@ -120,6 +121,30 @@ export const Standings = () => {
                   ? (classColorHex ?? highlightHex)
                   : highlightHex;
 
+            const mfrMap = new Map<string, { carId: number; count: number }>();
+            for (const s of classStandings) {
+              if (s.carId === undefined) continue;
+              const mfr =
+                CAR_ID_TO_CAR_MANUFACTURER[s.carId]?.manufacturer ?? 'unknown';
+              if (!mfrMap.has(mfr)) mfrMap.set(mfr, { carId: s.carId, count: 0 });
+              mfrMap.get(mfr)!.count += 1;
```
tariknz: Don't use non-null assertion please, do the additional check

Foribald:
Pushed \`64766d1\` with two fixes found while reviewing this PR's hook/assertion usage:

**Non-null assertion removed** — \`Standings.tsx\`'s new manufacturer-count computation had \`mfrMap.get(mfr)!.count += 1\`. Replaced the \`has()\`+\`set()\`+\`get()\` pattern with a single \`get()\` capturing the entry into a local, checked with a plain \`if (existing)\` before mutating — removes the assertion and the redundant extra map lookup.

**\`useIsSingleMake\` now gated by \`enabled\`** — this hook does a per-driver manufacturer scan, but was called unconditionally in both \`Standings.tsx\` and \`Relative.tsx\` even when \`hideIfSingleMake\` is off (the default), wasting the scan on every render for users who never enabled the setting. Added an \`enabled\` param that skips the scan inside the \`useMemo\` when off — the underlying \`useSessionDrivers\`/\`useWeekendInfoNumCarClasses\` subscriptions stay live regardless, so toggling the setting on mid-session recomputes correctly from current data on the very next render, not stale data. Added \`useIsSingleMake.spec.ts\` (no prior coverage existed) covering that exact toggle-on-mid-session case.

Verified with \`tsc --noEmit\`, \`eslint .\` (confirmed zero \`no-non-null-assertion\` violations repo-wide, cross-checked by intentionally introducing one in a throwaway file to confirm the rule actually fires), and the full suite (87 files / 1063 tests, 3 new) — all clean.

tariknz (approved):

PR #701, opened 2026-08-22
feat: lap graph that survives a full race, and readable standings columns

23Kev:
## Description

Reworks the Gantry Lap Graph so it aligns to general UX and functions as expected, and fixes a column overlap in the Gantry standings.

**The lap graph only recorded while its window was open.** Recording ran in the renderer and was gated on the Lap Graph tab being mounted, so opening the Gantry at hour three of a race showed the last few minutes, and closing the window lost everything. Recording now runs in the main process alongside the incident feed (`LapHistoryRuntime`), outside channel demand gating, and persists to disk keyed on `SubSessionID` (current race plus one previous). It keeps recording with the window closed and reloads after an app restart.

**Recorded gaps were wrong in exactly the situations endurance racing produces.** The old store wrote `driver.gap?.value ?? 0`, so every pit lap was recorded as a gap of zero — plotting as "briefly led the class". It also discarded the whole-laps-down count and had the sign stripped by `Math.abs`. The new `lap-history.snapshot` channel records raw lap crossings — lap, `SessionTime`, class position, and pit/off-track/lapped flags — and everything else is derived client-side. Nothing is defaulted, nothing is absolute-valued.

**The chart did not scale and ignored the theme.** It used a fixed 560x240 viewBox that letterboxed in a resizable window, hard-coded font families, pixel font sizes and rgba greys, and capped colour at 8 drivers with the other 52 collapsing into an opaque grey wash. It is now a two-canvas renderer (base plus overlay) sized by `ResizeObserver`, with axes, legend and readout kept in the DOM so the overlay theme's font, size and palette apply. Pointer movement redraws only the overlay.

Three y-axis modes: **Trace** (default), **Position** and **Gap**, with a lap window, brush overview, wheel zoom, pan and follow-live. Configurable under Gantry settings (`lapGraph`), with a config version bump from 1 to 2 and a migrator.

**Two bugs found while testing this in iRacing and fixed here:**

- *The chart went permanently empty after the race ended.* The channel bus drops its cached snapshot when the last subscriber unsubscribes, and `lap-history.snapshot` only publishes when a lap completes. Once the last car finished, switching Gantry tabs unsubscribed, dropped the cache, and no publish was ever due again. The runtime now resends the recorded race when a window resubscribes. Every other channel publishes continuously, so this only bites an event-rate channel. Regression test included, and it was confirmed to fail without the fix.
- *The race trace read as though even the leader was losing time.* The trace was anchored to the leader's start-line crossing, so the cost of the opening lap (86.18s against an 81.35s reference in the session checked) dropped the whole field by ~5s and held it there. It now anchors on the leader's first completed racing lap. Verified against a real saved race: the leader went from `0, -4.83, -4.28 ... -3.83` to `0, +0.5, +1.0 ... +0.6`. The opening lap is also excluded from the reference pace, though on that data it moved the median by only 0.04s — the anchor was the real cause.

**Standings column overlap.** The Best, Last and lap-delta columns were too narrow for a lap time and could be squeezed further by flex, so values ran together (`1:45.5301:47.443`). Best also carried `gap-px-1`, which is not a real Tailwind class, so it had no padding at all. Widths are now defined once and shared between the header and the rows so they cannot drift, and are sized in `ch` so they still fit when the overlay theme scales the font up.

Removed: `LapGapStore`, `LapGapStoreUpdater`, `LapGapChart` and their specs.

### Validation

Automated, all run on the final commit:

- `npm run lint` (tsc + eslint `--max-warnings 0`) — exit 0
- `vitest run` — 192 files, 1832 tests, exit 0
- `npm run test:replay:curated` — 36000 frames, 70 session revisions, 14 probes, exit 0

iRacing: the lap graph was exercised in multiple live races and in a spectated session. The post-race empty-chart bug above was found diagnosed from the app log and the persisted lap-history file, and the trace anchor fix was verified numerically against that saved race.

Known limitation: if `SessionNum` changes mid-`SubSession`, in-memory history is cleared and is not reloaded from disk, because reload is keyed on `SubSessionID`. This did not occur in the race checked.

### Before

image

### After

Screenshot 2026-08-22 062944

Lap graph: fills the window at any size, Trace / Position / Gap modes, class colours matching the standings pane, lap window with a brush overview and follow-live, and the recorded race still present after the flag and across tab switches. In Trace mode the class leader now sits flat near zero rather than several seconds below it.

Standings: Best, Last and the three lap-delta columns each sit in their own space with no overlap.

## Type of Change

- [x] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Added lap-history recording and session-based persistence.
  - Introduced interactive Gantry lap graphs with trace, position, and gap views.
  - Added zooming, panning, live-follow, hover details, lap navigation, class selection, and automatic car pinning.
  - Added settings for graph mode, visible lap range, and automatic pinning.
- **Bug Fixes**
  - Improved tooltip placement and standings-column sizing across display scales.
  - Preserved graph selections when switching views.
- **Refactor**
  - Replaced the legacy lap-gap data flow with the new lap-history system.


src/types/migrators/gantry.ts
```diff
@@ -0,0 +1,98 @@
+import { LAP_GRAPH_LAP_WINDOW_BOUNDS } from '../widgetConfigs';
+import type { LapGraphConfig, LapGraphYAxisMode } from '../widgetConfigs';
+
+/** Current schema version of the Gantry widget config. */
+export const GANTRY_CONFIG_VERSION = 2;
```
tariknz: @23Kev there's no need to create a migration as gantry is still not released I suggest just asking existing devs to nuke the config if needed.

tariknz (approved):

PR #709, opened 2026-08-29
fix: overlays sit empty when opened during a running session

23Kev:
## Description

Opening irdashies while a session is already running left the overlays empty for several seconds.

**Fix — running state is published as soon as telemetry arrives.**

`publishIRacingSDKEvents` seeded the running state from `sdk.sessionStatusOK` immediately after `ready()`, where it still reads `false`. The only correction was a 5-second poll, so every overlay opened during a live session was told the sim was down and waited for that poll. The telemetry loop already knew better — it sets `wasRunning = true` on its first frame — so it now publishes the change there. The 5s interval stays as the fallback that catches disconnects.

Measured on a packaged build, joining a live 72-car race:

| | before | after |
|---|---|---|
| running state published | 5.08s | **0.17s** |
| overlays visible | ~5s | ~4s |
| Gantry populated | ~19s | ~4s |

**Added — debug logging for the startup path, which had none.**

- `ProcessorHost`: first publish per channel, and separately the first publish that actually carries data. A channel's first publish is usually empty because the processor has just been activated by its first subscriber, so the two timestamps separate "channel silent" from "channel publishing but empty". Both markers reset on disconnect.
- `LapTimesStoreUpdater`: first snapshot, and first snapshot where any car has a lap time.

These log at `debug`, which the file transport dropped. `logger.ts` gains an `IRDASHIES_LOG_LEVEL` env override so the instrumentation can be captured on demand; the default is unchanged at `info`.

Known remaining issue, not addressed here: the standings widget still takes ~11s to show rows after its channel delivers a populated snapshot. The new logging is what localised it, and the next step is a marker inside `useDriverStandings`.

### Before

Overlays render empty on launch into a running session, then populate after several seconds.

### After

Overlays populate as soon as their data arrives. No visual design changes — timing only, so screenshots would be identical.

## Type of Change

- [x] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

---

Validation:

- `npx vitest run` — 200 files, 1934 tests pass. New test in `iracingSdkBridge.spec.ts` seeds `sessionStatusOK = false`, advances 100ms, and asserts the sequence is `[false, true]`; before the fix `true` only arrived at 5000ms.
- `npx tsc --noEmit` and `npx eslint` — clean.
- Verified in a live iRacing race from a packaged build of this branch, joining mid-session with 72 cars. Timings above are from that run's log.

## Summary by CodeRabbit

* **Bug Fixes**
  * Overlays now receive the simulator’s running-state update immediately when telemetry begins, rather than waiting for the next polling interval.

* **Improvements**
  * Added configurable file logging levels through the `IRDASHIES_LOG_LEVEL` setting.
  * Added diagnostic logging to help identify the first telemetry and lap-time data received, including after reconnects.

* **Tests**
  * Added coverage confirming running-state updates are published promptly when telemetry arrives.

### DIFF
--- a/src/app/bridge/iracingSdk/iracingSdkBridge.spec.ts
+++ b/src/app/bridge/iracingSdk/iracingSdkBridge.spec.ts
@@ -77,9 +77,11 @@ describe('publishIRacingSDKEvents session polling', () => {
     publishMessageToOverlay: vi.fn(),
   });
 
-  it('publishes the running state as soon as telemetry arrives', async () => {
+  it('publishes the running state as soon as the SDK produces data', async () => {
     // sessionStatusOK still reads false when the bridge seeds it, which is what
-    // left overlays believing the sim was down until the 5s poll caught up.
+    // left overlays believing the sim was down until the 5s poll caught up. A
+    // successful waitForData is the proof the sim is up, matching the existing
+    // wasRunning / lifecycle enter trigger; getTelemetry may still be null.
     mockSdkState.sessionStatusOK = false;
     const overlayManager = createOverlayManager();
 
--- a/src/app/bridge/iracingSdk/iracingSdkBridge.spec.ts
+++ b/src/app/bridge/iracingSdk/iracingSdkBridge.spec.ts
@@ -77,6 +77,28 @@ describe('publishIRacingSDKEvents session polling', () => {
     publishMessageToOverlay: vi.fn(),
   });
 
+  it('publishes the running state as soon as telemetry arrives', async () => {
+    // sessionStatusOK still reads false when the bridge seeds it, which is what
+    // left overlays believing the sim was down until the 5s poll caught up.
+    mockSdkState.sessionStatusOK = false;
+    const overlayManager = createOverlayManager();
+
+    const bridge = await publishIRacingSDKEvents(overlayManager as never);
+
+    // Well inside the 5s poll interval.
+    await vi.advanceTimersByTimeAsync(100);
+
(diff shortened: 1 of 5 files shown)


tariknz (dismissed):

tariknz:
The immediate running-state fix looks directionally sound, but I think there are two things worth addressing before merge:

1. `ProcessorHost.logFirstPublishes()` keeps calling `inspectSnapshot()` for scalar-only channels because `firstPopulatedPublishLogged` is never set when `shape.hasArrays === false`. That means repeated `Object.entries/filter/reduce` allocations in a publish path even when debug logging is disabled, which conflicts with the hot-path allocation guidance in R13.2. Mark non-array channels complete immediately; ideally also avoid shape inspection when debug logging is not enabled.

2. The new test says telemetry arrived, but the mocked `getTelemetry()` always returns `null`. The implementation therefore publishes `runningState: true` based on `waitForData()` alone, before verifying a telemetry object exists. Either move `publishRunningState(true)` inside `if (telemetry)`, or return telemetry from the test and explicitly document that successful `waitForData()` is the intended proof.

Minor related point: `LapTimesStoreUpdater` uses `.filter()` on every snapshot until a positive lap time appears, regardless of the active log level. A non-allocating count and/or debug-level gating would avoid diagnostic overhead potentially lasting several minutes.

The core bug fix itself looks low risk; my main concern is keeping the added instrumentation out of ongoing high-frequency work.

tariknz (approved):

PR #718, opened 2026-09-03
feat: set a layout per session type instead of reconfiguring between sessions

### LINKED ISSUE #608: [FR] Session-Specific Layout Configuration
Currently, layout configurations are applied globally across all sessions. This makes it difficult to tailor views for different session types within the same event.

I would like to request support for session-specific layout configurations, allowing different layouts to be assigned to sessions such as Practice, Qualifying, Warmup, and Race.

For example, I may want my standings table to display different columns depending on the session:

Qualifying: Position, Best Lap, Gap to Pole, Number of Laps
Race: Position, Interval, Gap to Leader, Pit Stops, Delta

Currently this is not possible without reconfiguration between session or switching profiles. I would like it to be a one time setup, and only start and forget to use afterwards.

CtrlAltDale:
Closes #608.

## Description

Layout configuration is global, so running different columns in qualifying and the race means reconfiguring between sessions or switching profiles by hand. The request asks for "a one time setup, and only start and forget to use afterwards".

Read literally, per-session variants of every widget setting would be a config explosion and a migration problem. But **profiles already carry a complete dashboard each**, and switching one already live-reloads the overlays without destroying windows. So this maps a profile to each session type and applies it as the event moves along — which means everything can differ per session, not just columns: widget positions, and which widgets exist at all.

**Nothing changes for an existing install.** A trigger with no profile mapped means "leave the profile alone", and an unrecognised session type gets no opinion rather than a guess.

### Spotting

Spotting gets its own trigger, because it is a *state* rather than a session type — it can happen inside any of them. It outranks the session type while the player is out of the car and hands back the moment they are driving again.

The outbound edge waits 20s, the inbound is immediate. Getting out of the car is not always deliberate — a tow, a quick garage visit, the pause during a driver swap — and throwing the whole layout around on every blip would be worse than switching a little late.

### Where it hooks in

Two derived events are added to `sessionLifecycle`, which already sees both halves of the data and is the single session event source:

- `onSessionTypeChange` resolves `SessionNum` against the session list in **either arrival order** (the number arrives on the telemetry tick, the types with session info) and fires on *type* changes only — so two consecutive practice sessions do not churn the layout.
- `onDrivingStateChange` mirrors the renderer's `useDrivingState`: in the car counts even when stationary in the pit box; the garage and replay playback do not.

The mapping crosses IPC, so it is rebuilt from an allowlist of trigger keys rather than trusted.

### Session types are validated against real data, not guessed

Checked against **205 recorded sessions**. iRacing emitted exactly seven distinct `SessionType` strings and all seven are handled:

| SessionType | Occurrences |
| --- | --- |
| `Practice` | 144 |
| `Race` | 100 |
| `Lone Qualify` | 61 |
| `Offline Testing` | 55 |
| `Open Qualify` | 19 |
| `Warmup` | 4 |
| `Lone Practice` | 2 |

Two findings worth recording, both captured in a comment in `sessionProfiles.ts`:

- **A Time Trial reports as `Lone Practice`** (with `SessionName: TIME TRIAL`), so it has its own trigger rather than silently landing on the practice profile.
- **Heat events run several sessions that all report `Race`**, separated only by `SessionSubType` (Heat / Consolation / Feature). They therefore share one profile, and moving between heats does not re-trigger a switch.

### On switching while driving

Deliberately **not** gated on the player being stationary. Gating would mean never getting the race layout during a race, since qualifying to race fires while you may already be rolling. The switch takes the same near-instant path as a manual profile change (`dashboardBridge.ts`: no destroy/recreate), though widgets that differ between the two profiles still have windows opened or closed — that is inherent to the feature. This was specifically watched for on the rig.

### Before

No such setting exists. Changing layout between sessions means switching profiles by hand, or reconfiguring widgets.

### After

A new **Switch Profile By Session** section under Settings, Profiles, with a row per session type and one for spotting below a divider:

```
Switch Profile By Session

  Practice             [ Default       v ]
  Time Trial           [ Do not switch v ]
  Open Qualifying      [ Qualifying    v ]
  Lone Qualifying      [ Do not switch v ]
  Warmup               [ Do not switch v ]
  Race                 [ Race          v ]
  Offline Testing      [ Do not switch v ]
  ----------------------------------------
  Spotting / watching  [ Spotting      v ]
```

Two Storybook stories cover it: `components/Settings/ProfileSettings` gives `Default` (how it looks before configuration) and `SessionMappingConfigured`.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

`defaultDashboard.ts` is untouched: this adds no widget and no widget config. The mapping is an app setting alongside `cycleProfiles` and `showProfileBanner`, and it seeds on first run rather than shipping a committed default.

## Validation

- **On the sim rig**, in iRacing, against a packaged build. Two problems were found there and fixed before this PR: the mapping started empty and looked inert (now seeded on first run), and the switcher was constructed before `publishDashboardUpdates` registered the listener that applies a switch — so a switch for the session already in progress emitted into an `EventEmitter` with no listener, silently leaving the stored profile and the visible overlays disagreeing.
- **Session-type strings** validated against 205 real captures, as above. `Warmup` in particular was a guess until it was confirmed verbatim in a recorded session.
- **Tests verified by mutation, not by passing.** 20 deliberate mutations across the switcher and the lifecycle — removing the already-current guard, removing the deleted-profile guard, breaking spotting precedence, removing the dwell and its cancellation, making an unmapped session inherit a default, letting the garage count as driving, firing on every tick instead of on the edge, and so on. Every one was caught by the test that names it. One early mutation was itself a no-op and was rewritten as a real one rather than counted.
- **Storybook** verified in a browser, which caught a fixture bug that jsdom did not: the story's bridge overrides were not reaching the component because `TelemetryDecorator` nests its own `DashboardProvider`.
- `npm test` — 1952 tests across 198 files, all passing.
- `npm run lint` — clean. `npm run build-storybook` — succeeds.

## Summary by CodeRabbit

- **New Features**
  - Added automatic profile switching for supported session types, including Practice, Qualifying, Race, Warmup, Time Trial, and Offline Testing.
  - Added **Switch Profile By Session** settings for configuring session and spotting profile mappings.
  - Profiles now switch when entering or leaving the car, with a brief delay before switching to the spotting profile.
  - Session profile mappings are saved and restored between launches.
  - Unknown or unconfigured session types leave the active profile unchanged.

### DIFF
--- a/src/app/sessionLifecycle/sessionLifecycle.spec.ts
+++ b/src/app/sessionLifecycle/sessionLifecycle.spec.ts
@@ -38,6 +38,25 @@ const makeEmptySession = (): Session =>
 const makeSessionWithMissingDrivers = (): Session =>
   ({ DriverInfo: {} }) as unknown as Session;
 
+const makeSessionList = (
+  sessions: { SessionNum: number; SessionType: string }[]
+): Session =>
+  ({
+    SessionInfo: { Sessions: sessions },
+    DriverInfo: { Drivers: [{ CarIdx: 0, CarIsPaceCar: 0, IsSpectator: 0 }] },
+  }) as unknown as Session;
+
+const makeDrivingTelemetry = (
+  sessionNum: number,
+  flags: Record<string, boolean>
+): Telemetry =>
+  ({
+    SessionNum: { value: [sessionNum] },
+    ...Object.fromEntries(
+      Object.entries(flags).map(([key, value]) => [key, { value: [value] }])
+    ),
+  }) as unknown as Telemetry;
+
 describe('sessionLifecycle', () => {
   let lifecycle: ReturnType<typeof createSessionLifecycle>;
 
@@ -299,6 +318,140 @@ describe('sessionLifecycle', () => {
       lifecycle._onSession(makeSession([0]));
 
       expect(goodSpy).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('session type', () => {
+    it('fires for the first session, because entering an event is the transition', () => {
+      const typeSpy = vi.fn();
+      lifecycle.onSessionTypeChange(typeSpy);
+
+      lifecycle._onSession(
+        makeSessionList([{ SessionNum: 0, SessionType: 'Practice' }])
+      );
+      lifecycle._onTelemetry(makeTelemetry(0));
+
+      expect(typeSpy).toHaveBeenCalledWith('Practice');
+    });
+
+    it('resolves when the session list arrives after the telemetry tick', () => {
+      const typeSpy = vi.fn();
+      lifecycle.onSessionTypeChange(typeSpy);
+
+      lifecycle._onTelemetry(makeTelemetry(2));
+      expect(typeSpy).not.toHaveBeenCalled();
+
+      lifecycle._onSession(
+        makeSessionList([{ SessionNum: 2, SessionType: 'Race' }])
+      );
+
+      expect(typeSpy).toHaveBeenCalledWith('Race');
+    });
+
+    it('does not fire when the session number changes but the type does not', () => {
+      const typeSpy = vi.fn();
+      lifecycle.onSessionTypeChange(typeSpy);
+      lifecycle._onSession(
+        makeSessionList([
+          { SessionNum: 0, SessionType: 'Practice' },
+          { SessionNum: 1, SessionType: 'Practice' },
+        ])
+      );
+
+      lifecycle._onTelemetry(makeTelemetry(0));
+      lifecycle._onTelemetry(makeTelemetry(1));
+
+      expect(typeSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('fires as the event moves from practice to qualifying to race', () => {
+      const typeSpy = vi.fn();
+      lifecycle.onSessionTypeChange(typeSpy);
+      lifecycle._onSession(
+        makeSessionList([
+          { SessionNum: 0, SessionType: 'Practice' },
+          { SessionNum: 1, SessionType: 'Open Qualify' },
+          { SessionNum: 2, SessionType: 'Race' },
+        ])
+      );
+
+      lifecycle._onTelemetry(makeTelemetry(0));
+      lifecycle._onTelemetry(makeTelemetry(1));
+      lifecycle._onTelemetry(makeTelemetry(2));
+
+      expect(typeSpy.mock.calls).toEqual([
+        ['Practice'],
+        ['Open Qualify'],
+        ['Race'],
+      ]);
+    });
+
+    it('starts clean after a disconnect', () => {
+      const typeSpy = vi.fn();
+      lifecycle.onSessionTypeChange(typeSpy);
+      lifecycle._onSession(
+        makeSessionList([{ SessionNum: 0, SessionType: 'Race' }])
+      );
+      lifecycle._onTelemetry(makeTelemetry(0));
+      typeSpy.mockClear();
+
+      lifecycle._onDisconnect();
+      lifecycle._onSession(
+        makeSessionList([{ SessionNum: 0, SessionType: 'Race' }])
+      );
+      lifecycle._onTelemetry(makeTelemetry(0));
+
+      expect(typeSpy).toHaveBeenCalledWith('Race');
+    });
+  });
+
+  describe('driving state', () => {
+    const drivingCases: [string, Record<string, boolean>, boolean][] = [
+      ['on track', { IsOnTrack: true }, true],
+      ['stationary in the pit box', { PlayerCarInPitStall: true }, true],
+      ['rolling down pit lane', { OnPitRoad: true }, true],
+      ['in the garage', { IsOnTrack: true, IsInGarage: true }, false],
+      ['watching a replay', { IsOnTrack: true, IsReplayPlaying: true }, false],
+      ['spotting for a team-mate', { IsOnTrack: false }, false],
+    ];
+
+    it.each(drivingCases)(
+      'reports %s as driving=%o -> %s',
+      (_label, flags, expected) => {
+        const drivingSpy = vi.fn();
+        lifecycle.onDrivingStateChange(drivingSpy);
+
+        lifecycle._onTelemetry(makeDrivingTelemetry(0, flags));
+
+        if (expected) {
+          expect(drivingSpy).toHaveBeenCalledWith(true);
+        } else {
+          // False is the starting assumption, so only a true->false edge fires.
+          expect(drivingSpy).not.toHaveBeenCalledWith(true);
+        }
+      }
+    );
+
+    it('fires only on the edge, not on every tick', () => {
+      const drivingSpy = vi.fn();
+      lifecycle.onDrivingStateChange(drivingSpy);
+
+      lifecycle._onTelemetry(makeDrivingTelemetry(0, { IsOnTrack: true }));
+      lifecycle._onTelemetry(makeDrivingTelemetry(0, { IsOnTrack: true }));
+      lifecycle._onTelemetry(makeDrivingTelemetry(0, { IsOnTrack: true }));
+
+      expect(drivingSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('fires both ways when the player hands the car over and takes it back', () => {
+      const drivingSpy = vi.fn();
+      lifecycle.onDrivingStateChange(drivingSpy);
+
+      lifecycle._onTelemetry(makeDrivingTelemetry(0, { IsOnTrack: true }));
+      lifecycle._onTelemetry(makeDrivingTelemetry(0, { IsOnTrack: false }));
+      lifecycle._onTelemetry(makeDrivingTelemetry(0, { IsOnTrack: true }));
+
+      expect(drivingSpy.mock.calls).toEqual([[true], [false], [true]]);
     });
   });
 });
--- a/src/app/sessionLifecycle/sessionLifecycle.spec.ts
+++ b/src/app/sessionLifecycle/sessionLifecycle.spec.ts
@@ -423,12 +423,13 @@ describe('sessionLifecycle', () => {
 
         lifecycle._onTelemetry(makeDrivingTelemetry(0, flags));
 
-        if (expected) {
-          expect(drivingSpy).toHaveBeenCalledWith(true);
-        } else {
-          // False is the starting assumption, so only a true->false edge fires.
-          expect(drivingSpy).not.toHaveBeenCalledWith(true);
-        }
+        // Asserted the same way for both outcomes. The driving state starts
+        // unknown rather than false, so the first frame is an edge either way
+        // and must fire exactly once with the state it found — a "did not fire
+        // with true" assertion would also pass if nothing fired at all, and the
+        // spotting dwell timer depends on that first false arriving.
+        expect(drivingSpy).toHaveBeenCalledTimes(1);
+        expect(drivingSpy).toHaveBeenCalledWith(expected);
       }
     );
 
--- a/src/app/sessionLifecycle/sessionLifecycle.spec.ts
+++ b/src/app/sessionLifecycle/sessionLifecycle.spec.ts
@@ -39,11 +39,15 @@ const makeSessionWithMissingDrivers = (): Session =>
   ({ DriverInfo: {} }) as unknown as Session;
 
 const makeSessionList = (
-  sessions: { SessionNum: number; SessionType: string }[]
+  sessions: { SessionNum: number; SessionType: string }[],
+  subSessionId?: number
 ): Session =>
   ({
     SessionInfo: { Sessions: sessions },
     DriverInfo: { Drivers: [{ CarIdx: 0, CarIsPaceCar: 0, IsSpectator: 0 }] },
+    ...(subSessionId === undefined
+      ? {}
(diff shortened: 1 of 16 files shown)


src/frontend/components/Settings/sections/ProfileSettings.tsx
```diff
@@ -25,14 +71,33 @@ export const ProfileSettings = () => {
...
+  ) => {
+    // An empty selection drops the key entirely rather than storing a blank,
+    // so "no profile" and "profile deleted" stay the same thing on disk.
+    const next = Object.fromEntries(
+      Object.entries({ ...sessionProfileMap, [triggerKey]: profileId }).filter(
+        ([, id]) => Boolean(id)
+      )
+    ) as SessionProfileMap;
+    await bridge.setSessionProfileMap?.(next);
+    setSessionProfileMap(next);
```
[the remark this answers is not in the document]
CtrlAltDale: Fixed in 84dd7ea1 — good catch, both halves were real.

**Stale write while loading.** The rows rendered enabled while `getSessionProfileMap` was still in flight, so choosing a profile in that window built a new map from an empty one and persisted it, wiping whatever was stored. The rows are now disabled until the fetch resolves.

**Lost selection on rapid edits.** Each edit derived the whole map from the one captured in that render, so two in quick succession both built on the same state and the second dropped the first. Edits now derive from a ref holding the newest map, and writes are chained so they reach storage in the order they were made rather than racing.

Verified in Storybook that the rows are disabled on first paint and enabled once the map arrives, with the stored mapping intact.

src/app/sessionLifecycle/sessionLifecycle.spec.ts
```diff
@@ -301,4 +320,138 @@ describe('sessionLifecycle', () => {
...
+        const drivingSpy = vi.fn();
+        lifecycle.onDrivingStateChange(drivingSpy);
+
+        lifecycle._onTelemetry(makeDrivingTelemetry(0, flags));
+
+        if (expected) {
+          expect(drivingSpy).toHaveBeenCalledWith(true);
+        } else {
+          // False is the starting assumption, so only a true->false edge fires.
+          expect(drivingSpy).not.toHaveBeenCalledWith(true);
```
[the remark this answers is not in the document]
CtrlAltDale: Fixed in 84dd7ea1. You are right, and the comment I had written to justify the assertion was wrong too.

`lastIsDriving` starts `undefined`, not `false`, so the first off-track frame *is* an edge and does fire `false` — which is exactly what starts the spotting dwell timer. My assertion only checked that the callback was not called with `true`, which passes just as well when nothing fires at all.

Both branches now assert the same way:

```ts
expect(drivingSpy).toHaveBeenCalledTimes(1);
expect(drivingSpy).toHaveBeenCalledWith(expected);
```

Worth noting for anyone reading later: I had mutation-tested this file and still missed it, because I mutated the production code rather than questioning the assertion itself. A vacuous negative survives that kind of check.

src/frontend/components/Settings/sections/ProfileSettings.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+export default meta;
+
+type Story = StoryObj<typeof ProfileSettings>;
+
+// No TelemetryDecorator here: it supplies its own DashboardProvider bound to
+// the shared mock bridge, and as the innermost decorator that provider would
+// shadow this one — leaving the story with the mock's single profile and no
+// session mapping. This section reads dashboard context only, so its provider
+// is all it needs.
+const withBridge = (map: SessionProfileMap): Story['decorators'] => [
```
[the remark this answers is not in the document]
CtrlAltDale: Not applying this one — I tried it and it breaks the story.

The ordering assumption is inverted for this setup. In this Storybook config the **first** decorator in the array ends up nearest the story, so adding `TelemetryDecorator()` before the custom one makes *its* `DashboardProvider` the innermost, and it shadows the story's provider rather than composing with it.

I applied the suggestion verbatim and loaded `SessionMappingConfigured` in a browser:

```
with TelemetryDecorator()     profiles: [Default]              rows: all "Don't switch"
without (current)             profiles: [Default, Qualifying,  rows: practice=Default,
                                         Race, Spotting]             openQualify=Qualifying,
                                                                     race=Race, spotting=Spotting
```

That is the exact bug this file was written to avoid — it is why the decorator was dropped in the first place, and the comment above `withBridge` records the reason.

Nothing is lost by leaving it out: this section reads dashboard context only, so its provider is all it needs. If it ever grows a dependency on telemetry context, the fix would be to make `TelemetryDecorator` accept an existing bridge rather than nesting a second provider.

src/frontend/components/Settings/sections/ProfileSettings.tsx
```diff
@@ -10,6 +21,45 @@ import {
...
+  disabled,
+  onChange,
+}: {
+  triggerKey: ProfileTriggerKey;
+  label: string;
+  profiles: DashboardProfile[];
+  value: string;
+  disabled: boolean;
+  onChange: (triggerKey: ProfileTriggerKey, profileId: string) => void;
+}) => (
```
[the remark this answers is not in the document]
CtrlAltDale: Half applied in 93a6ba92.

**Props type: done.** The inline type is now `SessionProfileRowProps`, which is the convention the guide asks for and was worth fixing.

**Directory and story: not applying.** That guideline describes widget components. Settings sections do not follow it, so moving this one would make it the exception rather than bring it into line. Counted in `Settings/sections`:

- 8 sections define sub-components inline exactly like this one (`BattleSettings`, `InputSettings`, `KeybindingsSettings`, `RelativeSettings`, `StandingsSettings`, `TachometerSettings`, `WeatherSettings`, `LayoutVisualizer`).
- 1 of 40 sections has a directory of its own (`FuelSettings`).
- 1 of 40 has a story (`KeybindingsSettings`).

`SessionProfileRow` is a private row renderer with no life outside this section. A directory and a story for it would be new structure rather than existing structure — happy to do it if a maintainer wants settings sections moved that way, but as a separate change covering all of them.

src/frontend/components/Settings/sections/ProfileSettings.tsx
```diff
@@ -25,14 +75,57 @@ export const ProfileSettings = () => {
...
+
+  // Every write sends the whole map, so it must be derived from the newest one
+  // rather than from a render's closure. Two edits in quick succession would
+  // otherwise both build on the same state and the second would drop the
+  // first's selection.
+  const sessionProfileMapRef = useRef<SessionProfileMap>({});
+  // Writes are chained so they reach storage in the order they were made. Two
+  // in flight at once can otherwise be persisted out of order, leaving the
+  // stored map disagreeing with what the page shows.
+  const sessionProfileWrites = useRef<Promise<unknown>>(Promise.resolve());
```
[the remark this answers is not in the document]
CtrlAltDale: Not applying this one, on consistency grounds.

`sessionProfileMap` sits next to two settings of exactly the same kind in the same component — `cycleProfiles` and `showProfileBanner`. All three are persisted configuration, all three are read once through the dashboard bridge on mount, all three are written back on change, and all three live in local `useState` today. Moving only the new one to context would leave three identical concerns implemented two different ways in one file, which is harder to follow than the status quo.

It also has a single consumer. The map is loaded, edited and saved entirely within this section; nothing else in the app reads it. The main process is the actual source of truth and already serialises access through the bridge.

The data-loss risks the earlier finding raised are real and are fixed in 84dd7ea1 — the rows are disabled until the fetch resolves, edits derive from a ref holding the newest map rather than a render closure, and writes are chained so they land in order. Context would not have prevented either of those; both were about ordering, not about where the state lived.

If the settings page should hold its configuration in context, I agree that is the better shape — but as one change that moves all three together, rather than one that splits them.

src/main.ts
```diff
@@ -399,6 +411,25 @@ app.on('ready', async () => {
...
+  // in would emit into an EventEmitter with no listener, silently leaving the
+  // stored profile and the visible overlays disagreeing.
+  //
+  // On a first run this points every session type at the profile already in
+  // use, so the settings page shows what the mapping is for. It is a no-op
+  // behaviourally: every type resolves to the active profile until one is
+  // changed.
+  initialiseSessionProfileMap(getCurrentProfileId());
+  sessionProfileSwitcher = createSessionProfileSwitcher({
+    lifecycle: getSessionLifecycle(),
```
tariknz, line 423: [P2] Seed the switcher with the current lifecycle state

The SDK is already publishing before this subscription is installed: iRacingSDKSetup starts its background loop, and startup then awaits other work before reaching this point. onSessionTypeChange/onDrivingStateChange only emit transitions and do not replay their current values to a new subscriber. If irDashies starts while iRacing is already in a session, the initial events can therefore be missed, leaving the mapped session profile unapplied (and the spotting timer unstarted) until another transition occurs.

Please initialize the switcher from the current lifecycle state after attaching its listeners, or provide a subscription that seeds that state. A regression test should emit the initial session/driving state before creating the switcher.

src/app/storage/appSettings.ts
```diff
@@ -18,3 +21,45 @@ export const getShowProfileBanner = (): boolean => {
...
+ */
+export const initialiseSessionProfileMap = (
+  currentProfileId: string
+): SessionProfileMap => {
+  const stored = readData<SessionProfileMap>(SESSION_PROFILE_MAP_KEY);
+  if (stored !== undefined) return stored;
+
+  const seeded: SessionProfileMap = {};
+  for (const key of SESSION_PROFILE_KEYS) {
+    seeded[key] = currentProfileId;
```
tariknz: [P2] Leave automatic profile switching disabled until configured

Seeding every session type with the current profile is only a no-op while that profile remains active. For example, first startup seeds profile A; the user later manually selects profile B without configuring this feature; the next practice/qualifying/race transition switches them back to A. This enables a behavior change for existing users despite the documented default of leaving their profile alone.

Please default to an empty mapping (or require an explicit opt-in before seeding it). Add coverage for a manual profile change after initialization followed by a session transition.

src/app/sessionLifecycle/sessionLifecycle.ts
```diff
@@ -58,6 +79,38 @@ export function createSessionLifecycle(): SessionLifecycle {
...
 
+  /**
+   * The session number arrives on the telemetry tick and the types arrive with
+   * session info, in no guaranteed order, so resolution is attempted after
+   * either one moves.
+   */
+  function resolveSessionType(): void {
+    if (lastSessionNum === -1) return;
+    const sessionType = sessionTypesByNum.get(lastSessionNum);
+    if (!sessionType || sessionType === lastSessionType) return;
```
23Kev, line 111: There is a bug here if you go from pre-race session practice (SessionType=Practice) to Warmup in the race session (SessionType=Practice). This means during the in race session practice the profile doesn't switch and stays on the Spotting/Watching profile.
According to Claude the fix could be.
Track the subsession in sessionLifecycle and clear the type state when it changes. In _onSession ([line 163] before building sessionTypesByNum:

```
let lastSubSessionId: string | undefined;   // with the other state, near line 59

const subSessionId =
  session?.WeekendInfo?.SubSessionID != null
    ? String(session.WeekendInfo.SubSessionID)
    : undefined;

if (subSessionId !== undefined && subSessionId !== lastSubSessionId) {
  lastSubSessionId = subSessionId;
  sessionTypesByNum.clear();   // session numbers are per-subsession
  lastSessionType = undefined; // so the new subsession's type fires even if the string matches
}
```

src/main.ts
CtrlAltDale: Fixed in b70eebf3. Confirmed: `onSessionTypeChange`/`onDrivingStateChange` only add to a callback set and never replay, so anything wired after `iRacingSDKSetup` misses what already resolved.

The lifecycle now exposes `getCurrentState()`, and the switcher seeds from it once after subscribing. Someone already out of the car at that point skips the spotting dwell — it exists to ride out brief hops out, and they have been out longer than the app has been running.

src/app/storage/appSettings.ts
CtrlAltDale: Fixed in b70eebf3 — agreed, and it breaks the claim in my own docblock. Seeding is only a no-op while the seeded profile stays active: seed A, pick B by hand without ever opening the feature, and the next session transition drags you back to A.

The map now starts empty. The discoverability that seeding bought is a "Use <profile> for all sessions" button on the settings page instead, so nothing is mapped until someone asks for it.

src/app/sessionLifecycle/sessionLifecycle.ts
CtrlAltDale: Thanks — real bug, fixed in b70eebf3, and your diagnosis was right. `sessionTypesByNum` and `lastSessionType` were only cleared on disconnect, and changing subsession does not disconnect the SDK, so a warmup reported as `Practice` was deduplicated against the practice session before it. Keyed on `SubSessionID` as you suggested; `incidentRuntime` and `lapHistoryRuntime` already key on it the same way. I also drop the stale session number, otherwise it gets looked up in the new event's table and briefly announces the wrong session.

One addition: that alone would not have unstuck you. The switcher also never learned the session type it started in, so it held no session key at all and spotting was the only trigger that could ever apply. Both are fixed. If your practice and warmup were in the *same* subsession (no qualifying between them) let me know — the ID would not change there, and the seed is what covers that case.

tariknz (dismissed):

tariknz (approved):

PR #719, opened 2026-09-05
feat: tell lap graph lines apart, and plot position for the whole race

23Kev:
## Description

Two fixes for the Gantry lap graph.

**Line colours.** Every car in a class was drawn in the same colour, so you could not tell one line from another. Each driver now gets a colour and dash pattern based on their qualifying position, fixed for the whole race. Ten colours across four dash patterns cover 40 cars.

The colours were checked with a palette validator instead of picked by eye. Cars that qualified near each other get the furthest apart colours, and the set holds up for red-green colour blindness. Colour on its own cannot separate ten lines, so the dash pattern, the driver list swatch and the hover readout all help identify a car.

**Position chart.** It stopped part way through a race. It read `classPosition` from each lap crossing, but iRacing leaves `CarIdxClassPosition` at 0 most of the time. In a recorded 20 lap race it was 0 on 81% of crossings, and on every crossing after lap 6.

Positions now come from the order cars crossed the line on each lap. Lap number and session time are always recorded, and crossing order is correct for lapped cars too. Replayed against the same race file, the chart went from 6 laps and 23 cars to 21 laps and 24 cars. Drivers with no usable position could not be selected in the driver list before, and now can.

**Driver list.** Restyled to match the main standings list, and moved into its own component folder. Each row shows a swatch of that driver's line.

**Qualifying results.** Moved into a shared `useQualifyingGrid` selector. The heat race fallback used to build a new array on every render, which rebuilt the whole standings tree each time. Standings and Relative get that fix as well.

### Before

All lines in a class drawn in one colour. In Position mode the chart ended at lap 6 of a 20 lap race, with several drivers greyed out in the driver list.

### After

Screenshot 2026-09-05 215939

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

### DIFF
--- a/src/frontend/components/Gantry/components/LapGraph/LapGraphCanvas.spec.tsx
+++ b/src/frontend/components/Gantry/components/LapGraph/LapGraphCanvas.spec.tsx
@@ -2,11 +2,19 @@ import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
 import { LapGraphCanvas } from './LapGraphCanvas';
 import type { LapGraphSeries } from './LapGraphCanvas';
+import { identityForGridSlot } from './lapGraphPalette';
 
 interface FakeContext {
   strokes: number;
   clears: number;
   fills: number;
+  /** Dash array passed to the most recent `setLineDash` call before a stroke. */
+  lastDash: readonly number[];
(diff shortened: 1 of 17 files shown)


CtrlAltDale (approved):
Ran through the Storybook stories with a full field and it looks really nice — the patterns plus the tooltip make it easy to follow a driver even when the colours get close together. Nice job.

PR #720, opened 2026-09-05
fix: the Gantry is empty after its window is minimised during a race

23Kev:
## Description

The Gantry sits empty after its window has been minimised for a race. Incidents, standings and the lap graph were all affected, for three separate reasons.

**Event channels were dropped while hidden.** `ChannelBus.publish()` skipped any window that was not visible and threw the payload away. That is fine for snapshot channels, because a hidden window is re-sent the latest snapshot when it comes back. Event channels are never re-sent, so anything dropped is lost for good. In one 32 minute race, 98 incidents were detected and written to disk but only the last 9 reached the list. Three channels are events: `session.lifecycle`, `raceControl.incidents` and `raceControl.sessionId`. All are low volume, so sending them to a hidden window costs almost nothing. They are no longer gated on visibility. Snapshot channels keep the existing behaviour. This also fixes `raceControl.sessionId`. A session change never reached a minimised Gantry, so the incident list was never reloaded for the new session.

**Session data never reached the window.** The drivers list arrives on the legacy `sessionData` stream. It is published only when iRacing changes its session string, and only to visible subscribed windows. On Windows a minimised window reports itself as not visible, so a Gantry minimised for the race never received it, and the standings and lap graph had no field to draw. There was no recovery path either: nothing seeded a renderer when it subscribed, and nothing re-sent the data when the window returned. It is now seeded on subscribe, re-sent on `show` and `restore`, and the window fires the ready callbacks when it loads, so it also picks up the running state when opened mid-session.

**Lap time history stopped recording.** `ProcessorHost` counts visible subscribers as demand, so the lap times processor was destroyed while its only subscriber was hidden. That lost the history behind the Gantry's L-3, L-2 and L-1 columns, which stayed blank for three more laps after a restore. A processor can now opt in to running while its windows are hidden, and it is pushed the current snapshot when one becomes visible. Only `lap-times.snapshot` opts in. Delivery is still gated on visibility, so hidden windows still receive nothing.

Only minimising the window caused this. Closing the Gantry and reopening it reloads everything from disk.

The four event delivery tests were checked by putting the old behaviour back. All four fail with the visibility check restored.

Tested in iRacing by leaving the Gantry hidden for a full race, then restoring it.

### Before

Hidden for the race, then restored. The incident feed held only the last few incidents, the standings list was empty, and the lap graph read "Waiting for the grid".

### After

Incidents, standings and the lap graph all fill in as soon as the window is shown, with the lap delta columns populated straight away.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

## Summary by CodeRabbit

- **Behavior Updates**
  - Event-based updates continue to be delivered while a renderer or view is hidden, including after hide/show transitions.
  - Snapshot-based updates pause while hidden and resume when visibility is restored.
  - Subscriptions created while hidden remain active and receive future event updates.
  - Selected background processing continues while all subscribed views are hidden.
  - Newly shown or restored views receive the latest available session data without waiting for a new update.
  - Current session data is seeded when a renderer subscribes.

### DIFF
--- a/site/src/utils/previewChannelRuntime.ts
+++ b/site/src/utils/previewChannelRuntime.ts
@@ -77,8 +77,8 @@ export function createPreviewChannelRuntime(
     isDestroyed: () => disposed,
     // Always active. Gating this on `document.visibilityState` looks like a
     // free CPU saving but is a trap: a subscription created while the document
-    // reports hidden never fires `notifySubscriberCount`, so ProcessorHost
-    // never activates the processor and the preview stays permanently empty.
+    // reports hidden counts as no visible demand, so ProcessorHost never
+    // activates the processor and the preview stays permanently empty.
     // A document can report hidden while still painting (occluded tab,
     // automation, some embeds). The saving is redundant anyway — browsers
     // already throttle timers and pause rAF for genuinely backgrounded tabs.
--- a/src/app/bridge/channelBridge.spec.ts
+++ b/src/app/bridge/channelBridge.spec.ts
@@ -369,4 +369,91 @@ describe('ChannelBus', () => {
     expect(target.send).toHaveBeenCalledOnce();
     expect(target.send).toHaveBeenCalledWith(CHANNEL_DELIVERY, 'snapshot', 7);
   });
+
+  it('reports the registered count alongside the active count', () => {
+    const { bus } = createBus();
+    const target = createTarget();
+    const counts: [number, number][] = [];
+    bus.onSubscriberCountChanged((channel, activeCount, registeredCount) => {
+      if (channel === 'snapshot') counts.push([activeCount, registeredCount]);
+    });
+
+    bus.subscribe(target, 'snapshot', 10);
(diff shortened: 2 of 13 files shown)


CtrlAltDale (dismissed):

CtrlAltDale (approved):
Approving. Three separate causes for one symptom, each with its own fix — and the 98-detected/9-shown figure makes the event-channel one hard to argue with.

The snapshot-versus-event distinction is the right call: a hidden window can be re-seeded from the latest snapshot, but a dropped event is gone for good. Good that only the three event channels are ungated and snapshot delivery is unchanged.

The `processWhileHidden` opt-in is the part I looked hardest at, since demand-gating decides whether a processor runs at all. Opt-in with a single user, and delivery still gated on visibility, keeps that contained.

Also good to see the event tests checked by putting the old behaviour back rather than just watching them pass.

PR #722, opened 2026-09-05
fix: overlays hide behind the sim and the settings window

CtrlAltDale:
## Description

Since #716, overlays disappear behind the sim and behind the settings window. On my rig, every time:

- Enter edit mode (F6), click and hold a widget: **every widget vanishes**. The drag still works — the widget moves where you put it, you just cannot see anything.
- On release they do not come back. The next F6 does nothing; a second F6 puts you back *into* edit mode; only a third leaves edit mode with the widgets visible.
- Click into the simulator: **all widgets disappear**.
- Select irDashies from the taskbar: the **settings window comes back without the widgets**.

That last symptom is the one that identifies the cause.

### What is actually going on

The overlays and the settings window use a deliberate two-level window band, and `toggleLockOverlays` says so in a comment:

```ts
// Raise settings window to layer 2 during edit mode so it appears above
// overlay windows (layer 1); revert to normal layering when not editing.
this.currentSettingsWindow.setAlwaysOnTop(true, 'screen-saver', 2);
```

That only works if the overlays are actually on layer 1. #716 removed the call that put them there:

```diff
-    if (this.overlayAlwaysOnTop) {
-      browserWindow.setAlwaysOnTop(true, 'screen-saver', 1);
-    }
```

and left only the constructor flag, which has no level:

```ts
alwaysOnTop: this.overlayAlwaysOnTop,
```

A plain `alwaysOnTop` window sits in a lower band than a `screen-saver` one. So the settings window still climbs to layer 2 while the overlays no longer climb to layer 1 — the settings window now outranks them outright, and so does a fullscreen sim. "Taskbar brings back the settings screen but not the widgets" is exactly what that looks like.

Worth noting: this call was itself the fix from #286 ("remove hack to force on top"), which replaced a `setInterval` that re-asserted always-on-top every five seconds. Removing it brought back the problem that interval had existed to paper over.

### The change

Restore the call, guarded on the setting so #716's actual intent — honour the user's choice rather than forcing always-on-top — is preserved:

```ts
if (this.overlayAlwaysOnTop) {
  browserWindow.setAlwaysOnTop(true, 'screen-saver', 1);
}
```

Three lines. The constructor flag from #716 stays exactly as it is.

### A test had to change, and I want to flag it rather than bury it

#716 added:

```ts
expect(createdWindows[0].setAlwaysOnTop).not.toHaveBeenCalled();
```

That assertion codifies the regression, so it could not stay. It is replaced by two tests that pin the actual contract:

- always-on-top on: `setAlwaysOnTop(true, 'screen-saver', 1)` is called, with a comment explaining why the level matters and what breaks without it.
- always-on-top off: `setAlwaysOnTop` is not called at all, so the user's choice is still honoured.

The `alwaysOnTop=%s` constructor test from #716 is kept unchanged apart from dropping that one line.

### Before

Widgets vanish on click-and-hold in edit mode and do not return; clicking into the sim hides them all; the taskbar restores the settings window without them.

### After

Widgets stay visible while dragging in edit mode, survive clicking into the sim, and come back with the app from the taskbar.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

No Storybook story: this is main-process window layering, which Storybook cannot exercise.

## Validation

**Isolated with a control build before writing any code.** I first suspected a widget I was working on locally. To check, I built the current `main` with that widget removed entirely and ran the same steps on the rig — every symptom reproduced on a build containing none of my code. That ruled my work out and pointed at this window.

**Confirmed fixed in iRacing**, Toyota GR86 at Virginia, practice. Same steps as the reproduction above: widgets now stay visible through a drag, survive clicking into the sim, and return with the app from the taskbar.

**Unit tests, verified by mutation** rather than by passing. Four deliberate breakages, each caught by the test that names it:

| Mutation | Caught |
| --- | --- |
| Revert to the post-#716 state (no call at all) | yes |
| Use level 2 instead of level 1 | yes |
| Drop the level argument, `setAlwaysOnTop(true)` | yes |
| Ignore the user's setting and always call it | yes |

`npm test` — 1914 tests across 196 files, all passing. `npm run lint` — clean.

**One caveat, stated because it is load-bearing:** I have verified the behaviour empirically on Windows, but not that Electron's `level` argument is documented to affect Windows z-ordering the way this code assumes. The evidence for the diagnosis is the codebase's own stated intent, the history of #286, and the symptoms matching the layering exactly — plus the fix working on the rig. If a maintainer knows a more precise mechanism, I would rather the comment said that.

## Summary by CodeRabbit

- **New Features**
  - Overlay windows configured to stay on top now maintain the intended stacking order, appearing above regular windows while remaining below the settings window during edit mode.

- **Bug Fixes**
  - The “always on top” overlay setting is now applied consistently when creating overlay windows.

### DIFF
--- a/src/app/overlayManager.spec.ts
+++ b/src/app/overlayManager.spec.ts
@@ -159,9 +159,45 @@ describe('OverlayManager display windows', () => {
 
       expect(createdWindows).toHaveLength(1);
       expect(createdWindows[0].options.alwaysOnTop).toBe(overlayAlwaysOnTop);
-      expect(createdWindows[0].setAlwaysOnTop).not.toHaveBeenCalled();
     }
   );
+
+  it('puts an always-on-top overlay in the same band as the settings window', () => {
+    // The constructor flag alone is not enough. toggleLockOverlays raises the
+    // settings window to 'screen-saver' 2 during edit mode so it sits above the
+    // overlays, which only works if the overlays are at 'screen-saver' 1. With
+    // the level left off, the settings window and a fullscreen sim both end up
+    // above the overlays instead, and the widgets vanish.
+    const manager = new OverlayManager();
+
+    manager.createOverlays(
+      {
+        widgets: [],
+        generalSettings: { overlayAlwaysOnTop: true },
+      } as DashboardLayout,
+      { createSettingsWindow: false }
+    );
+
+    expect(createdWindows[0].setAlwaysOnTop).toHaveBeenCalledWith(
+      true,
+      'screen-saver',
+      1
+    );
+  });
+
+  it('leaves the level alone when the user has turned always-on-top off', () => {
+    const manager = new OverlayManager();
+
+    manager.createOverlays(
+      {
+        widgets: [],
+        generalSettings: { overlayAlwaysOnTop: false },
+      } as DashboardLayout,
+      { createSettingsWindow: false }
+    );
+
+    expect(createdWindows[0].setAlwaysOnTop).not.toHaveBeenCalled();
+  });
 
   it('shows a ready display overlay without activating it', () => {
     const manager = new OverlayManager();
--- a/src/app/overlayManager.ts
+++ b/src/app/overlayManager.ts
@@ -273,6 +273,19 @@ export class OverlayManager {
       visibleOnFullScreen: true,
     });
 
+    // The overlays and the settings window use a deliberate two-level scheme:
+    // overlays at 'screen-saver' 1, and the settings window raised to
+    // 'screen-saver' 2 during edit mode so it sits above them (see
+    // toggleLockOverlays). The constructor's `alwaysOnTop` flag alone puts the
+    // window in a lower band than either, so the settings window — and a
+    // fullscreen sim — end up above the overlays instead of below.
+    //
+    // Guarded on the setting so it still honours the user's choice, which is
+    // what the constructor flag above was changed to do.
+    if (this.overlayAlwaysOnTop) {
+      browserWindow.setAlwaysOnTop(true, 'screen-saver', 1);
+    }
+
     if (MAIN_WINDOW_VITE_DEV_SERVER_URL) {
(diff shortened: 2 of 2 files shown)


23Kev:
Pulled this down and re-ran the checks locally: `npm run lint` clean, full suite green (200 files, 1944 tests). CodeRabbit raised nothing. The change looks correct and minimal, and the diff is an exact restoration of what #716 dropped.

**On your open question about `level` and Windows z-ordering.** It is load-bearing, and here is the mechanism. From Electron 44, `shell/browser/native_window_views.cc`:

```cpp
// On macOS the window is placed behind the Dock for the following levels.
// Re-use the same names on Windows to make it easier for the user.
static constexpr auto levels = base::MakeFixedFlatSet<std::string_view>(
    {"floating", "torn-off-menu", "modal-panel", "main-menu", "status"});
behind_task_bar_ = levels.contains(level);
```

The constructor's `alwaysOnTop: true` resolves to level `"floating"` (the default in `native_window.h`), which is in that set, so the overlay gets pushed behind the taskbar band. `'screen-saver'` is not in the set, so it stays above. That accounts for the symptoms you listed, and it is why restoring the call fixes them.

One nuance worth recording: `relativeLevel`, the trailing `1` here and the `2` on the settings window, is never read in the Windows implementation. So on Windows the load-bearing part is the `'screen-saver'` string rather than the number, and the relative two-level band is macOS-only.

Two things I noticed that this PR does not introduce. Both are pre-#716 behaviour returning, so I would raise them separately rather than hold this up:

- Edit mode expands the overlay to `display.bounds` rather than `workArea` and turns click-through off. Above the taskbar band that means the overlay covers the taskbar and takes its clicks until you leave edit mode.
- `overlayAlwaysOnTop` is only read in `createOverlays`, so a renderer crash recovery or a newly attached display recreates the window with the startup value rather than the current setting.

Good write-up, and the mutation table is a useful way to show the tests actually hold.

23Kev (approved):
Changes approved

PR #724, opened 2026-09-06
fix: app starts with Windows despite the setting being off

CtrlAltDale:
## Description

Reported by a user: irDashies starts with Windows on every boot even though **Start on system startup** is switched off in the profile they are using. They traced it themselves — Explorer was launching `irdashies.exe` from `HKCU\Software\Microsoft\Windows\CurrentVersion\Run`, value name `com.squirrel.irdashies.irdashies`. No scheduled task, service, or Startup folder shortcut involved.

### What is happening

Autostart is stored per profile, but applied from the profile literally named `default`:

```ts
public setupAutoStart(): void {
  const dashboard = getDashboard('default');   // not the profile in use
  app.setLoginItemSettings({
    openAtLogin: dashboard?.generalSettings?.enableAutoStart ?? false,
  });
}
```

So two code paths disagree:

1. Toggling the setting off calls `bridge.setAutoStart(false)`, which applies to the OS **immediately**, and writes `enableAutoStart: false` into the **active** profile.
2. The next launch calls `setupAutoStart()`, which reads **`default`**, finds `true`, and **re-creates the Run entry**.

The settings screen keeps showing "off" throughout, because the UI reads the active profile. The reporter had `enableAutoStart: false` on their active profile (GT3) and `true` on `default`, so it turned itself back on every single launch.

### Two more instances of the same bug

`getDashboard('default')` appears three times in `overlayManager.ts`, and all three are read outside a dashboard update:

| Function | Consequence |
| --- | --- |
| `setupAutoStart()` | the reported bug |
| `setupHardwareAcceleration()` | hardware acceleration follows `default`, not the profile in use |
| `shouldCloseToTray()` | close-to-tray follows `default` |

All three now read the active profile through one helper. Leaving the other two would have meant knowingly shipping two instances of a bug I had just diagnosed.

A fourth `getDashboard('default')` in `deleteProfile` is **left alone** — it runs immediately after deliberately switching to `default`, so it is correct.

### Autostart is also re-applied on profile switch

Previously the Run entry described whichever profile the app started on until the next launch. Since these settings differ per profile, switching profile now re-applies it, so the OS state and the settings screen cannot drift apart while the app is running.

### Note on the wider design

The reporter's first suggestion was to make autostart a **global** application setting rather than a per-profile one, which I agree is the better shape — autostart, hardware acceleration and close-to-tray are per-machine concerns, and `storage/appSettings.ts` already exists for exactly that. That needs a migration for existing configs and a decision about where the toggles live in the UI, so it is not in this PR. This is the contained fix for the reported bug; happy to follow up on the design separately if you want it.

## Screenshots

No visual change — this is main-process startup behaviour.

### Before

Autostart shown as off in the active profile, but a `Run` registry entry is re-created on every launch and Windows starts the app at login.

### After

The registry entry matches what the settings screen shows, on startup and after a profile switch.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

Not tested in iRacing: this is login-item and window behaviour in the main process, with nothing to observe in a live session.

## Validation

**These three functions had no test coverage at all**, which is how this survived. The new tests pin the reported scenario directly rather than testing the mechanism abstractly:

- autostart **off** in the active profile and **on** in `default` must not enable it — this is the reported bug, and it fails on `main`
- autostart on in the active profile still enables it
- `getDashboard` is called with the active profile id and **not** with `'default'`
- a profile that does not mention `enableAutoStart` must not opt the user in
- hardware acceleration follows the active profile, both ways

**Verified by mutation, not by passing.** Three deliberate breakages:

| Mutation | Caught by |
| --- | --- |
| Revert autostart to reading `default` | "does not enable autostart from another profile" |
| Revert hardware acceleration to reading `default` | "takes hardware acceleration from the active profile too" |
| Default autostart to **on** when the setting is absent | "leaves autostart off when the profile does not mention it" |

The third mutation **survived the first pass** — my tests always set the value explicitly, so the `?? false` fallback was unpinned. Added a test for the unset case, and it is caught now. Worth mentioning because an absent setting silently enabling autostart is exactly the class of bug this PR is fixing.

`npm test` — 1920 tests across 196 files, all passing. `npm run lint` — clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Profile-specific settings for hardware acceleration, autostart, and close-to-tray behavior now follow the currently active profile instead of the default profile.
  - Switching profiles now refreshes the Windows autostart configuration to match the newly selected profile.

### DIFF
--- a/src/app/bridge/dashboard/dashboardBridge.ts
+++ b/src/app/bridge/dashboard/dashboardBridge.ts
@@ -256,6 +256,10 @@ export async function publishDashboardUpdates(
     clearTimeout(hideTimer);
     clearTimeout(settleTimer);
     lastProfileId = profileId;
+    // Autostart lives in the dashboard, so it differs between profiles. Re-apply
+    // it here or the Windows Run entry keeps reflecting the profile the app
+    // started on, while the settings screen shows the one now in use.
+    overlayManager.setupAutoStart();
     const name = getProfile(profileId)?.name ?? '';
     // Banner is cosmetic; the hide/reveal still runs to mask the resize.
     const showBanner = getShowProfileBannerStorage();
--- a/src/app/overlayManager.spec.ts
+++ b/src/app/overlayManager.spec.ts
@@ -55,6 +55,7 @@ vi.mock('electron', () => ({
   app: {
     getVersion: () => '0.0.0',
     disableHardwareAcceleration: vi.fn(),
+    setLoginItemSettings: vi.fn(),
     commandLine: { appendSwitch: vi.fn() },
   },
   BrowserWindow: FakeBrowserWindow,
@@ -74,7 +75,10 @@ vi.mock('electron', () => ({
 }));
 
 vi.mock('./storage/storage', () => ({ readData: vi.fn(), writeData: vi.fn() }));
-vi.mock('./storage/dashboards', () => ({ getDashboard: vi.fn() }));
+vi.mock('./storage/dashboards', () => ({
+  getDashboard: vi.fn(),
+  getCurrentProfileId: vi.fn(() => 'default'),
+}));
 vi.mock('./storage/chromiumFlags', () => ({
   getChromiumFlags: vi.fn(() => ({})),
   parseCustomSwitches: vi.fn(() => []),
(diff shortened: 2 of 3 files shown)


23Kev:
Reviewed — no issues with the fix itself. The diagnosis and the change both look right, and I confirmed the three converted reads leave no `getDashboard('default')` behind.

One thing worth a release note: **some users will lose autostart after this.**

`defaultDashboard.generalSettings.enableAutoStart` is `false`, and profile creation merges those defaults into every profile. So anyone who currently gets autostart because `default` has it `true`, while actually running a different profile, will find it stops.

That is the fix working as intended — their settings screen already showed it as off — but to them it will look like a regression, and it may generate "autostart stopped working" reports. Worth calling out in the release notes so it is expected.

23Kev (approved):
Review and approved, merging

PR #726, opened 2026-09-07
feat: keep pit laps out of the lap history graph

CtrlAltDale:
## Description

Requested by a user practising ovals: pit laps make the lap history graph hard to read.

A pit lap is ten or more seconds slower than a green one, so it stretches the y-axis and drags the average. The laps you are actually comparing get squashed into a flat band a few pixels tall between the spikes, which is exactly when a tight oval spread matters most.

This adds an optional **Hide Pitted Laps** toggle that leaves those laps out of both the list and the graph.

### Detecting a pit lap

A lap is marked pitted if `PlayerTrackSurface` reads `InPitStall` or `ApproachingPits` at any point during it.

It is **latched for the lap** rather than sampled when the lap is scored — by the time the line is crossed the surface reads `OnTrack` again, so sampling at the crossing would miss every stop.

That also covers the out-lap without special-casing it: an out-lap starts in the box, so the surface already reads `InPitStall` when it begins. Both halves of a stop drop out together, which is what the reported screenshot shows — the spikes come in pairs.

### Filtering happens before the count

Hiding pit laps still shows the configured number of laps. Filtering after the slice would quietly shorten the list — ask for ten laps after two stops and get seven.

Defaults to **off**, so no existing display changes.

### Also fixes a pre-existing bug in the same check

The off-track half of the dirty-lap test compared the surface against a hardcoded `4`:

```ts
const TRACK_SURFACE_OFF_TRACK = 4;
```

`TrackLocation` runs `NotInWorld = -1` to `OnTrack = 3`, so `4` is not a value it can produce and the comparison could never succeed. Only the incident half of the check worked, so going off without picking up an incident left the lap looking clean.

Both surface tests are now small exported predicates with tests over every enum member. The failure mode here is silence rather than an error, which is how this survived — so the mapping is worth pinning.

This is a separate defect from the feature and easy to split out if you would rather it landed on its own.

## Screenshots

Two Storybook stories added on the same ten-lap practice stint with two stops, so the difference is visible without a sim: `widgets/LapTimeLog` → **Pit Laps Shown (default)** and **Pit Laps Hidden**.

### Before

`AVG 0:34.820`. Two spikes dominate the chart; the green laps sit flat along the bottom and the lap-to-lap variation is unreadable.

### After

`AVG 0:31.287`. The average now reflects green-lap pace, the y-axis rescales, and the variation between laps fills the chart.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

Not yet tested in iRacing — the pit-surface detection is covered by unit tests and the display by Storybook, but the latch has not been watched through a real stop. Happy to run a practice stint before this merges if you would rather have that first.

## Validation

**Storybook, before and after**, on the same stint: the average moves from `0:34.820` to `0:31.287` and the y-axis rescales, which is the actual complaint being fixed.

**Filter tests, verified by mutation** rather than by passing. Three deliberate breakages, each caught by the test that names it:

| Mutation | Caught by |
| --- | --- |
| Never filter, so the toggle does nothing | "leaves out the pit laps when the toggle is on" |
| Always filter, ignoring the toggle | "shows every lap by default, so nothing changes for an existing setup" |
| Filter after the slice, so the list silently shortens | "still fills the configured number of laps when some are hidden" |

**Surface predicates** are asserted against every member of `TrackLocation`, including an explicit case that `4` is not off-track — the exact value that made the original bug invisible.

`npm test` — 2000 tests across 199 files, all passing. `npm run lint` — clean. `npm run build-storybook` — succeeds.

### One thing worth knowing about the diff

`demoData.ts` shows about 33 lines of change that are **not** mine: the file was not prettier-formatted, and the pre-commit hook normalised it when I touched it. My change to that file is the two-line `pitted` field; the rest is quote and whitespace normalisation. Happy to revert that hunk if you would rather keep it out.

## Summary by CodeRabbit

- **New Features**
  - Added a **Hide Pitted Laps** option to lap history settings.
  - Lap history can now exclude laps involving pit stops while retaining laps around the stop.
  - Added visual examples showing pitted laps displayed or hidden.

- **Bug Fixes**
  - Improved pit-stop and off-track detection for more accurate lap history records.
  - Filtering now occurs before the history display limit is applied.

### DIFF
--- a/src/frontend/components/LapTimeLog/hooks/useLapTimeLog.ts
+++ b/src/frontend/components/LapTimeLog/hooks/useLapTimeLog.ts
@@ -30,7 +30,16 @@ export const isPitSurface = (surface: number): boolean =>
 
 export const isOffTrackSurface = (surface: number): boolean =>
   surface === TrackLocation.OffTrack;
-const MAX_HISTORY_ENTRIES = 20;
+/**
+ * The retained history has to outrun the display, not match it. "Number Of Laps
+ * To Show" goes up to 20, and with pitted laps hidden the display filters this
+ * list before taking that many — so a cap of 20 made the setting unsatisfiable
+ * the moment a stop happened: the clean laps that could have filled the gap had
+ * already been dropped here. 60 covers the maximum display count even on short
+ * fuel runs where two laps in three are in-laps or out-laps, and the entries are
+ * five numbers each.
+ */
+const MAX_HISTORY_ENTRIES = 60;
 const FREEZE_TIME = 5;
 
 export const useLapTimeLog = () => {
--- a/src/frontend/components/LapTimeLog/hooks/useLapTimeLog.ts
+++ b/src/frontend/components/LapTimeLog/hooks/useLapTimeLog.ts
@@ -5,10 +5,31 @@ import {
   useSessionStore,
   useDriverCarIdx,
 } from '@irdashies/context';
+import { TrackLocation } from '@irdashies/types';
 import type { LapEntry } from '../demoData';
 import { useLapTimeLogSettings } from './useLapTimeLogSettings';
 
-const TRACK_SURFACE_OFF_TRACK = 4;
+/**
+ * A lap counts as pitted if the car was in the pit lane or its box at any point
+ * during it. That covers the in-lap and the out-lap without special-casing
+ * either: an out-lap starts in the stall, so the surface already reads
+ * InPitStall when it begins.
+ */
+const PIT_SURFACES: readonly number[] = [
+  TrackLocation.InPitStall,
+  TrackLocation.ApproachingPits,
+];
+
+/**
+ * Exported so the mapping onto TrackLocation is testable. It is worth pinning:
+ * this file previously compared the surface against a hardcoded 4, which is not
+ * a value the enum can take, so the off-track check never fired.
+ */
+export const isPitSurface = (surface: number): boolean =>
+  PIT_SURFACES.includes(surface);
+
+export const isOffTrackSurface = (surface: number): boolean =>
+  surface === TrackLocation.OffTrack;
 const MAX_HISTORY_ENTRIES = 20;
 const FREEZE_TIME = 5;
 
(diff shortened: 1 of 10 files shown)


src/frontend/components/LapTimeLog/hooks/useLapTimeLog.ts
```diff
@@ -194,6 +225,7 @@ export const useLapTimeLog = () => {
           ? lastLapTime - referenceAtStartOfLap.current
           : 0,
       dirty: isDirty,
+      pitted: isPitted,
     };
     setHistory((prev) => [newEntry, ...prev].slice(0, MAX_HISTORY_ENTRIES));
```
tariknz, line 239: [P2] Retain enough clean laps before applying the display count

The display now filters before slicing, but this hook has already capped the source history at 20 total laps. With 20 laps selected and two pit laps among the latest 20, the widget can only show 18, even if older clean laps exist: those have already been discarded here. The new display-only test supplies the full history directly, so it cannot catch this production limit.

Please keep a bounded history that can supply the configured number of qualifying laps when filtering is enabled, and add a hook-to-display regression covering more than 20 completed laps with pit laps in the retained window.

CtrlAltDale: Fixed in 12e06766. You were right, and the arithmetic is worse than stated: the count dropdown offers up to 20 and the cap was 20, so with pitted laps hidden a single stop made the maximum setting permanently unsatisfiable.

Retained history raised to 60 so it outruns the display rather than matching it. The regression drives the real hook through a 40 lap stint and renders its output through the real display — at the old cap it yields 16 clean laps for a display asking for 20.

src/frontend/components/LapTimeLog/hooks/useLapTimeLog.ts
```diff
@@ -194,6 +225,7 @@ export const useLapTimeLog = () => {
           ? lastLapTime - referenceAtStartOfLap.current
           : 0,
       dirty: isDirty,
+      pitted: isPitted,
```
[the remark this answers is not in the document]
CtrlAltDale: Not applying. The race is real but needs pit entry and lap scoring in the same frame, and the suggested patch trades it for a worse failure: at tracks where pit entry sits just past the line, `isPitted || isPitSurface(playerTrackSurface)` marks the flat-out lap you just finished as pitted instead of the in-lap.

The existing latch already covers the realistic geometry — `ApproachingPits` is seen during the lap, well before it is scored.

tariknz (approved):

PR #731, opened 2026-09-12
feat: add the Lap Trace widget

Foribald:
## Description

Adds a **Lap Trace** widget. It plots a saved reference lap against the one you are driving, indexed by position on track rather than by time, so the two line up wherever the car is. The reference is drawn washed out behind a bright ghost of the current lap.

**Reference laps come from three sources**

- **Personal best** — recorded automatically per track and car. Only a clean lap is promoted, and it is timed from iRacing's own `LapLastLapTime` rather than the line crossing, so the stored time matches the sim's.
- **`.ibt` telemetry import** — the main process stream-scans the file for its fastest valid lap, so a multi-gigabyte file never crosses the IPC boundary. Raw pedal channels are preferred where the file has them, matching what the live recorder stores.
- **Garage 61 CSV import** — for a lap someone else drove. The export has no time column, so per-sample timing is reconstructed by integrating speed over the GPS positions.

**On top of the trace**

- A **brake-point countdown**: three beeps and a distinct tone, and/or a bar strip on any edge of the widget. Tones are synthesised rather than bundled, so there is no asset to resolve out of the asar and nothing for a future CSP to block. Each tone is configurable, and the cues can be routed to a chosen Windows playback device so they can play on a headset while the sim stays on speakers.
- A **Last Corner panel** comparing the corner just finished against the reference: time delta, how many metres earlier or later you braked, and the minimum speed carried through. Corner names come from the bundled Lovely Sim Racing track data, which does not cover every layout — where it is missing the panel is simply absent, which the Help tab explains.
- Gear labels, throttle and brake application markers, and an ABS highlight on the brake trace.

Settings are tabbed, with a **Help** tab covering both import paths, where iRacing writes telemetry, and the widget's limitations.

**Review pass**

A second commit hardens the feature against a self-review of the first:

- The `.ibt` importer now requires `PlayerCarMyIncidentCount` and rejects any candidate lap whose count moves, so the "fastest clean lap" it promises is one. It also requires a start/finish crossing to both open and close a lap, so the part-laps at either end of a recording cannot win on elapsed time, and it validates every header offset and count against the file size before allocating.
- Asynchronous reads in the store are tagged with a generation, so a read that resolves after a newer one cannot apply a previous session's best time or overwrite the reference source the driver just picked.
- The pending lap's time baseline comes from the frame before the crossing and is bound to `LapCompleted`, so a time iRacing publishes in the crossing frame is not mistaken for the baseline. An incident landing on that same frame now keeps the lap out of the best slot.
- Lap trace storage loads asynchronously behind one shared promise, because first access can happen mid-session when the widget is enabled or Settings is opened; and shutdown drains the write queue rather than writing synchronously alongside a write already in flight.
- Gantry is kept out of the normal overlay, which it could appear in alongside its own window.
- A brake cue point is re-armed once the car is no longer sitting on it, so stopping on one no longer silences it for the rest of the session.

A third commit covers a second round of the same review:

- A lap waiting for its official time is let go at the next crossing, before that frame's `LapLastLapTime` is looked at. It could otherwise be finalized with the following lap's time, storing one lap's trace under a time it never set.
- An incident first reported in the sample that opens the next lap is charged to the lap that earned it, rather than becoming the new candidate's baseline and leaving a dirty lap eligible as the fastest clean one.
- The recorder re-initializes when the store has been emptied under it. The sim disconnect clears the store while the recorder's cached session identity survives, so reconnecting to the same track, car and session left it silently idle.
- The browser/OBS renderer mounts the recorder, which is what allocates and feeds the active lap: without it that view drew no trace however many laps were driven. It has no bridge behind it, so an import source there now says imported laps are shown in the desktop overlay instead of showing the recording message.

**Validation**

- `npm test` — 2624 tests across 233 files, all passing. The feature ships its own tests, including the sliding-window geometry across the start/finish seam, brake-cue latching, corner attribution, the `.ibt` binary reader against a synthetic file builder, and the Garage 61 CSV parser.
- `npm run lint` — clean (`tsc --noEmit` and `eslint . --max-warnings 0`).
- Storybook stories for the widget, the countdown bars and the Last Corner panel, driven by a synthetic lap so they render without telemetry.
- Driven in iRacing.

### Before

No equivalent existed. There was no way to compare the lap being driven against a saved one.

### After

laptrace (4)
laptrace (10)
laptrace (12)
laptrace (14)

configuration options:
laptrace config 1
laptrace config 2
laptrace config 3
laptrace config 4
laptrace config 5
laptrace config 6

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist


tariknz:
Hey this sounds awesomne @Foribald, do you mind posting screenshots or videos do demo this?

Foribald:
> Hey this sounds awesomne @Foribald, do you mind posting screenshots or videos do demo this?

Hi, sorry for the delay, I was trying to get coderabbitai to review it. I have put screenshots.

I took the free trial but it doesn't see the PR as i am not owner of the repo (I think)

Foribald:
@coderabbitai full review

tariknz:
> > Hey this sounds awesomne @Foribald, do you mind posting screenshots or videos do demo this?
> 
> Hi, sorry for the delay, I was trying to get coderabbitai to review it. I have put screenshots.
> 
> I took the free trial but it doesn't see the PR as i am not owner of the repo (I think)

No problem I will trigger it from codex

src/app/ibt/ibtBestLap.ts
```diff
@@ -0,0 +1,223 @@
...
+ * the current candidate and the best — so an endurance file of any length costs
+ * the same as a single lap (the plan's memory-bounding requirement).
+ *
+ * The scanner is pure logic over samples: ibtImport.ts streams the file and
+ * feeds it, keeping this unit-testable against synthetic sample sequences.
+ */
+import type { IbtLapSamples } from '@irdashies/types';
+import { IbtImportError } from './ibtErrors';
+
+export type { IbtLapSamples };
```
tariknz, line 16: **[P1] Validate incidents when selecting an imported lap** — `IbtSample` does not carry `PlayerCarMyIncidentCount`, and the importer does not require or decode that channel. An incident lap can therefore be selected as the fastest “clean” lap, contrary to the importer’s user-facing promise. Include the incident count in the sample and reject candidates whose count changes.

src/frontend/context/LapTraceStore/LapTraceStore.tsx
```diff
@@ -0,0 +1,962 @@
...
+    // whichever source the user has chosen to display. Nothing may be promoted
+    // until this resolves, or the first lap of the session would overwrite a
+    // faster stored one simply by finishing before the read did.
+    if (!bridge) {
+      // No persistence at all, so there is no stored lap to protect.
+      set({ bestLapLoaded: true });
+      return;
+    }
+    try {
+      const stored = await bridge.getLapTrace(trackId, carPath, 'best');
```
tariknz, line 523: **[P1] Ignore stale initialization results** — this asynchronous read can resolve after a newer `initialize()` call has switched the store to a different track/car. Its `set()` then marks the current session loaded and applies the old session’s best time. Associate the request with the initialized identity (or cancel/ignore stale promises) before applying the result.

src/frontend/context/LapTraceStore/LapTraceStore.tsx
```diff
@@ -0,0 +1,962 @@
...
+          isOnTrack && !onPitRoad,
+          incidentCount,
+          true
+        );
+      }
+    }
+
+    if (
+      lap.isCleanLap &&
+      (onPitRoad || !isOnTrack || incidentCount > lap.incidentCountAtStart)
```
tariknz, line 717: **[P1] Ignore stale reference-source reads** — switching sources while this promise is pending allows an older request to resolve last and overwrite the newer `referenceSource`, `referenceLap`, and `referenceError`. Track a request generation/source before applying the result.

src/frontend/context/shared/useWidgetsForThisDisplay.ts
```diff
@@ -0,0 +1,57 @@
...
+ * display bounds of its own and should show the lot.
+ */
+export const useWidgetsForThisDisplay = (
+  browser = false
+): DashboardWidget[] => {
+  const { currentDashboard, containerBoundsInfo } = useDashboard();
+
+  return useMemo(() => {
+    const enabled =
+      currentDashboard?.widgets.filter((widget) => widget.enabled) ?? [];
```
tariknz, line 54: **[P2] Keep Gantry out of the normal overlay** — the previous `OverlayContainer` explicitly excluded the separately-rendered `gantry` widget. This shared filter now returns every enabled widget, so enabling Gantry can render it in both the dedicated Gantry window and the normal overlay.

src/app/ibt/ibtImport.ts
```diff
@@ -0,0 +1,251 @@
...
+      header.bufOffset <= 0
+    ) {
+      throw new IbtImportError('bad-signature', 'Not a valid .ibt file');
+    }
+
+    const sub = parseDiskSubHeader(prefix);
+
+    const varHeaderBuf = await readRegion(
+      handle,
+      header.varHeaderOffset,
```
tariknz, line 223: **[P2] Bound `numVars` before allocating** — this value comes directly from the file header and is multiplied into a `Buffer.alloc()` size before checking that the header region fits in the file or that the count is reasonable. A malformed `.ibt` can force a very large allocation. Validate offsets, counts, and region bounds against `size` first.

src/frontend/domain/lapTrace/brakeCueLatch.ts
```diff
@@ -0,0 +1,408 @@
...
+
+  if (Number.isNaN(state.targetM)) {
+    const index = nextEventIndex(
+      cuePointsM,
+      cuePointCount,
+      carDistanceM,
+      trackLengthM,
+      BRAKE_CUE_LOOKAHEAD_M
+    );
+    if (index < 0) {
```
tariknz, line 301: **[P2] Clear a passed target when stationary** — when the car is stopped on or just beyond a cue point, `nextEventIndex()` finds no point ahead and this branch returns without releasing the current target. The latch can remain stuck until movement happens in a favorable direction, preventing later cues. Detect the passed target and release/re-arm it even when speed is below the cue threshold.

src/frontend/domain/lapTrace/garage61CsvImport.ts
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * Converts a Garage 61 CSV lap export into a LapTraceRecord.
+ *
+ * A Garage 61 export carries no numeric iRacing TrackID/CarPath — only
+ * human-readable names in its filename — and this app has no lookup table
+ * from those names to real IDs. So the imported lap is not matched against
+ * the live session at all: it is stored under a fixed sentinel key and shown
+ * whenever 'garage61' is the selected reference source, regardless of which
+ * car/track is actually loaded. (Known limitation: if the live session is a
```
tariknz, line 14: **[P2] Do not silently compare unrelated imported laps** — the sentinel key and blank `trackConfigName` deliberately bypass identity checks, so a Garage 61 lap from another circuit (or an `.ibt` from another car) is shown as the current reference after rescaling. Either validate the imported identity against the live session or make the mismatch explicit in the widget before plotting.

src/frontend/context/LapTraceStore/LapTraceStore.tsx
```diff
@@ -0,0 +1,962 @@
...
+        // copied out immediately) and wait for its official LapLastLapTime.
+        // The placeholder time is patched in on finalise. Any still-unresolved
+        // pending lap is dropped here — after this crossing LapLastLapTime
+        // moves to a different lap, so the old one can no longer be trusted.
+        if (lap.isCleanLap) {
+          const record = snapshotActiveLap(lap, state, -1);
+          set({
+            pendingBest: {
+              record,
+              lastLapTimeAtBoundary: lastLapTime,
```
tariknz: **[P2] Associate the official lap time with the completed lap** — if `LapLastLapTime` changes in the same sample as the crossing, this stores the new value as the pending baseline. The completed lap then waits for another change and can be discarded or receive the following lap’s time. Retain the previous sample’s time and completed-lap identity when finalizing.

src/frontend/context/LapTraceStore/LapTraceStore.tsx
```diff
@@ -0,0 +1,962 @@
...
+      if (lap.lastTrackedPct > LAP_END_PCT && pct < LAP_START_PCT) {
+        // A clean lap just crossed the line: capture it now (its samples are
+        // copied out immediately) and wait for its official LapLastLapTime.
+        // The placeholder time is patched in on finalise. Any still-unresolved
+        // pending lap is dropped here — after this crossing LapLastLapTime
+        // moves to a different lap, so the old one can no longer be trusted.
+        if (lap.isCleanLap) {
+          const record = snapshotActiveLap(lap, state, -1);
+          set({
+            pendingBest: {
```
tariknz, line 689: **[P2] Check crossing-frame incidents before snapshotting** — the crossing frame can contain the first incident increment while `lap.isCleanLap` still reflects the previous frame. This snapshots the lap as clean and resets the incident baseline before the normal cleanliness check runs. Compare the crossing-frame count against the finishing lap’s baseline before creating `pendingBest`.

src/app/ibt/ibtBestLap.ts
```diff
@@ -0,0 +1,223 @@
...
+    }
+    return this.best;
+  }
+
+  private finalize(c: Candidate): void {
+    if (c.invalid) return;
+
+    const n = c.pct.length;
+    if (n < MIN_LAP_SAMPLES) return;
+    if (c.minPct > LAP_START_PCT || c.maxPct < LAP_END_PCT) return;
```
tariknz, line 227: **[P2] Require both lap crossings for an imported candidate** — `finish()` finalizes the trailing candidate, and the 90% coverage check accepts a recording that stops at 91% (or starts after 10%). Its elapsed time excludes part of the track and can beat every genuinely completed lap. Track explicit start and finish crossings and require both.

src/app/storage/lapTraces.ts
```diff
@@ -0,0 +1,256 @@
...
+  }
+};
+
+const flushSync = (): void => {
+  if (!cache) return;
+  if (writeTimer) {
+    clearTimeout(writeTimer);
+    writeTimer = null;
+  }
+  try {
```
tariknz: **[P2] Drain asynchronous writes before shutdown** — this synchronous write bypasses `writeInFlight`. If `flushAsync()` is already writing, both operations can write the same file concurrently and the older write can overwrite the final cache. Make shutdown flushing asynchronous, await the existing queue, then write the latest cache.

src/app/storage/lapTraces.ts
```diff
@@ -0,0 +1,256 @@
...
+
+/**
+ * Lazy-load the file into the in-memory cache on first access. Synchronous
+ * by design — this runs once at startup, which is permitted by R6.1.
+ *
+ * Records from another schema version are dropped here rather than migrated:
+ * the format has only ever shipped on a feature branch, and every reader
+ * downstream assumes the current shape.
+ */
+const loadCache = (): Map<string, LapTraceRecord> => {
```
tariknz: **[P2] Load the cache asynchronously on first access** — lap-trace IPC handlers can first access this cache when the user enables the widget or opens Settings mid-session, so this synchronous read can block the Electron main process and telemetry polling at runtime. Use a shared asynchronous initialization promise.

src/frontend/domain/lapTrace/garage61CsvImport.ts
Foribald: This would have to wait for garage61 api integration. CSV file name is supposed to have the track name and layout but it's inconsistent with iracing names. With api, I will have access to iracing track id and will implement this feature.

Foribald:
> > > Hey this sounds awesomne @Foribald, do you mind posting screenshots or videos do demo this?
> > 
> > 
> > Hi, sorry for the delay, I was trying to get coderabbitai to review it. I have put screenshots.
> > I took the free trial but it doesn't see the PR as i am not owner of the repo (I think)
> 
> No problem I will trigger it from codex

Please rerun it on this commit

src/frontend/context/LapTraceStore/LapTraceStore.tsx
```diff
@@ -0,0 +1,1035 @@
...
+      // so the fresh time is the first one that differs from what was showing
+      // when the lap crossed. Telemetry-only: a lap whose time never differs
+      // is left pending and dropped at the next crossing rather than
+      // mis-timed.
+      const pending = state.pendingBest;
+      // LapCompleted can lag the line by a tick, so the lap waiting here is
+      // either the number seen at the crossing or the one after it. Beyond
+      // that, a further lap has finished and LapLastLapTime has moved on to
+      // describe that one instead — this lap can no longer be timed.
+      if (pending && lapCompleted > pending.lapCompletedAtBoundary + 1) {
```
tariknz: **[P2] Drop the unresolved lap before accepting the next crossing's time**

The previous-sample baseline fixes the original same-frame case, but the completed-lap guard still permits the following lap's time. If LapLastLapTime stays at 90 when lap 1 finishes with LapCompleted=1, that trace remains pending. At the next crossing, LapCompleted=2 and LapLastLapTime=85 pass this guard, and finalization runs before crossing detection, saving lap 1's samples as 85 seconds. I reproduced this on the updated head with distinct throttle traces for the two laps. Associate the time with the completed lap explicitly, or discard the unresolved pending lap before processing the next crossing's time.

src/app/ibt/ibtBestLap.ts
```diff
@@ -0,0 +1,259 @@
...
+    const started = this.current !== null && this.current.pct.length > 0;
+    const lapChanged =
+      this.current !== null && sample.lap !== this.current.lapNumber;
+
+    if (this.current === null) {
+      // The file's first candidate: wherever the recording happens to have
+      // started, which is rarely the line.
+      this.current = this.newCandidate(sample.lap, false);
+    } else if (started && (lapChanged || pctWrap)) {
+      this.finalize(this.current, true);
```
tariknz, line 152: **[P2] Check the finishing candidate's incidents before finalizing it**

The added incident channel rejects increments within a candidate, but an increment first delivered on the finishing crossing is checked only after this finalization and becomes the next candidate's baseline. The just-finished incident lap is therefore still eligible as the fastest clean reference. A regression check with an incident increment at the crossing selects the dirty 45-second lap over a subsequent clean 60-second lap. Compare the crossing sample's incident count with the finishing candidate's baseline before calling finalize().

src/frontend/context/LapTraceStore/useLapTraceRecorder.ts
```diff
@@ -0,0 +1,200 @@
...
+      const trackConfigName = session.WeekendInfo?.TrackConfigName ?? '';
+      const subSessionId = session.WeekendInfo?.SubSessionID ?? -1;
+
+      const lengthStr = session.WeekendInfo?.TrackLength;
+      const [val, unit] = lengthStr?.split(' ') ?? [];
+      const trackLengthM =
+        unit === 'km' ? parseFloat(val) * 1000 : parseFloat(val);
+      if (!Number.isFinite(trackLengthM) || trackLengthM <= 0) return;
+
+      if (
```
tariknz, line 110: **[P2] Reset the cached session identity on disconnect**

useResetOnDisconnect clears the lap-trace store, but this hook's sessionRef survives and onSession returns immediately for a null session. On reconnect to the same track/car/subsession, this identity comparison skips bootstrap. The reset channel sample also leaves sessionNum=-1, so the first reconnect sample does not trigger bootstrap either. I reproduced the disconnect/reset/reconnect sequence on the updated head: activeLap remains null and recording never resumes until the recorder remounts or the session identity changes. Clear the cached identity on disconnect or reinitialize when the store has been reset.

src/frontend/App.tsx
```diff
@@ -74,13 +75,18 @@ const GantryApp = () => {
...
-      <ProfileSwitchOverlay>
-        <ThemeManager>
-          <OverlayContainer />
-        </ThemeManager>
-      </ProfileSwitchOverlay>
-    </HideUIWrapper>
+    <>
+      {/* Outside the hide wrapper: Alt+H unmounts everything inside it, and
+          unmounting the recorder abandons the lap being driven. */}
+      <LapTraceUpdater />
```
tariknz, line 81: **[P2] Initialize Lap Trace in the browser/OBS dashboard too**

LapTraceUpdater is mounted only in the Electron OverlayApp. The browser entry mounts RendererDataProviders and DashboardView, which renders the enabled Lap Trace widget but never mounts its recorder or initializes its renderer-local lap-trace store. The browser entry also provides no lapTraceBridge for loading a saved reference. Consequently the browser/OBS widget remains on 'Drive a clean lap to record a reference' even while laps are driven or the Electron overlay has a reference. Wire the browser renderer to the required recording/reference data as well.

Foribald:
@tariknz I hope it's ok now

tariknz (approved):

tariknz:
I'm excited for this. I might clean up some of the UI but will keep you looped in. Thanks heaps!

Foribald:
> I'm excited for this. I might clean up some of the UI but will keep you looped in. Thanks heaps!

sure! as you see fit!

PR #737, opened 2026-09-13
feat: add optional toggle to disable the built-in web server

pureyang:
## Description

Adds an optional `generalSettings.enableWebServer` toggle that lets users turn off the built-in component/OBS web server. **It defaults to `true`, so behaviour is unchanged for every existing install** — the key is absent from stored dashboards and `?? true` preserves today's behaviour.

Motivation: `startComponentServer()` runs unconditionally on every launch, so users who never stream or record still pay for an HTTP + WebSocket server they will never connect to. Two costs for that group:

- **Port availability.** The server takes port 3000 and silently falls back through 3001–3005. Those are the default ports for a lot of local dev tooling — Next.js, Create React App, Rails, Grafana and plenty more all want 3000/3001. If you sim race on the same machine you develop on, irDashies quietly occupies the port your dev server wanted, and because the fallback is silent it is not obvious which process took it.
- **System utilisation.** A listening socket, the WebSocket bridge, and the per-connection work behind it are all running for people who get nothing from them. Sim racing is already a machine-under-load situation; letting users trim what they do not use is worth a toggle.

Neither is a problem for anyone who actually uses browser sources — this simply lets the people who do not opt out. Default stays on, so streamers notice nothing.

The gate returns before `http.createServer`, so when disabled the app creates no server object and binds no socket at all — not a listener that 404s.

```ts
const webServerEnabled = dashboard?.generalSettings?.enableWebServer ?? true;
if (!webServerEnabled) {
  logger.info('Component server disabled via generalSettings.enableWebServer - not listening on any port.');
  return;
}
```

`ProfileSettings.tsx` swaps the OBS Browser Source URL panel for a short explainer when the server is off, so the UI never advertises a URL that cannot work.

One note on the diff: the `lint-staged` pre-commit hook reformatted an unrelated `fontWeight` union in `dashboardLayout.ts` onto a single line. That is Prettier's doing, not mine — it looks like a Prettier version difference from whenever that file was last formatted. Happy to drop that hunk if you would rather keep the diff strictly additive.

### How this was tested

Not tested in an iRacing session (see checklist). Verified instead by running the real `startComponentServer` under Electron against a real `config.json`, and then against a packaged Squirrel build on Windows:

| `enableWebServer` | Listening sockets | Log |
| --- | --- | --- |
| `false` | none across all 6 processes | `Component server disabled via generalSettings.enableWebServer - not listening on any port.` |
| `true` | `[::1]:3000`, `GET /health` → 200 | `Component server running on http://localhost:3000` |
| absent (existing installs) | `[::1]:3000` | unchanged from today |

Same binary in both directions, so the disabled result is the flag working rather than a silent startup failure. `npm run lint` and `npm test` pass locally (1913/1913).

### Before

Settings → General ends with **Enable Network Access**, then **Hardware Acceleration**. The web server always runs; there is no way to turn it off.

### After

A new **Enable Web Server** toggle sits immediately above **Enable Network Access**, defaulted on:

> **Enable Web Server**
> Runs the built-in web server that exposes your overlays as browser sources for OBS. When disabled, the app listens on no ports at all — freeing port 3000 — and the OBS browser source URLs stop working. Leave enabled if you stream or record. (requires restart)

With it off, the **OBS Browser Source** panel in Settings → Profiles is replaced by a note pointing back at the toggle.

## Type of Change

- [ ] Performance improvement
- [ ] Documentation update
- [ ] Dependency update

## Checklist

Notes on the unchecked boxes, so they are not silently skipped:

- **Discord** — not discussed. Happy to take this to Discord first if you would prefer that.
- **iRacing** — the change is in the Electron main process startup path and has no telemetry or SDK involvement, so it was verified against a packaged build rather than a live session. Say the word if you still want an in-sim run.
- **Storybook** — no new visual component; this adds a toggle to `GeneralSettings.tsx`, which has no existing story (`KeybindingsSettings` is the only storied section). Glad to add one if you want coverage there.
- **README** — the README documents the web server under *Built-in Web Server*. I left it alone to keep the diff focused, but I am happy to add a sentence noting the toggle.

## Pre-PR Checklist (per `docs/ARCHITECTURE_RULES.md`)

```
[x] N1 — no new fs.*Sync outside startup (verified: no fs.*Sync added)
[x] N2 — no new frontend → src/app/ imports (verified: none added)
[x] N3 — no new cross-widget imports
[x] N4 — no new ipcMain handlers
[-] R2.1/R2.2 — N/A, no telemetry involved
[-] R3.1 — N/A, no new stores
[-] R4.1 — N/A, no new bridges
[-] R6.1 — N/A, no new storage code (reuses existing generalSettings persistence)
[-] R7.1 — N/A, no new widget
[x] R8.1 — additive setting: type updated in dashboardLayout.ts and default added to
    defaultDashboard.ts; deepMergeConfig fills it for existing dashboards. No breaking
    change, so no migration required (R8.2 N/A).
[-] R10.1 — N/A, no native code
[x] R11.1 — logging uses the project logger with a literal level name
[-] R13.1 — N/A, startup path, not hot-path
[x] R14.1 — new componentServer.spec.ts covers default-on, explicit-on, explicit-off,
    and off-wins-over-enableNetworkAccess
[-] Storybook — no new visual component (see note above)
[x] Architecture impact: none. No layer, bridge, store, or telemetry changes. The only
    behavioural change is an early return in startComponentServer() guarded by a setting
    that defaults to the current behaviour. Not hot-path, so no perf numbers.
```

## Summary by CodeRabbit

- **New Features**
  - Added a desktop setting to enable or disable the built-in web server.
  - The web server is enabled by default.
  - Disabling the web server frees port 3000 and makes OBS browser-source URLs unavailable.
  - Settings now explain the impact of disabling the server.
  - OBS browser-source settings display a notice when the web server is disabled and provide guidance for re-enabling it.
- **Tests**
  - Added coverage for enabled, disabled, and default web-server startup behavior.


tariknz (approved):
Brilliant, thank you
Are you on the discord?

